PR #37, opened 2015-04-26
Modifications on test_logging and test_shells

sklerder:
Modification on test_logging in order to check correctly subdirectories in /var/log containing a number in subdirectory's name (LOGG-2150).

Modification on test_shells to check 'umask' values in '/etc/profile' and '/etc/bash.bashrc' (SHLL-6240 and SHLL-6250)


mboelen:
Items have been implemented in separated changes, so no merge will occur in this case.

PR #40, opened 2015-04-27
Add detection of chef-client and chef-server as an automation tool

tas50:
Detects chef-server installed using the Omnibus package, chef-client installed via the omnibus package, and chef-client in chef-dk with the omnibus package.


tas50:
Any chance this can get merged @mboelen  ?

mboelen:
Hi Tim,

Thanks for the input. It couldn't be merged in this way, as it would be introducing a lot of "foreign" binary paths searches. I have rewritten your input so that it should be detecting Chef. Can you have a look if it works correctly for you?

mboelen:
Closing this item. In case the current patch is not working, let us know via the tracker.

PR #42, opened 2015-04-27
OWNER?ID fails on FreeBSD

rsmith-nl:
On FreeBSD, the -n flag to 'ls' only works in combination with -l.
So the following code returns *nothing*;

    # ls -n include/consts | awk -F" " '{ print $3 }'

However, if we change it to ‘ln -ln’, it works;

    # ls -ln include/consts | awk -F" " '{ print $3 }'
    0


rsmith-nl:
Doesn't work on Linux.

PR #43, opened 2015-04-28
Add test for 'ipfw' firewall on FreeBSD.

rsmith-nl:
Tested on FreeBSD 10.1.


mboelen:
Thanks for the pull request. It has committed and merged.
Minor changes: changed exception ID and using IPFW (instead of ipfw).

PR #51, opened 2015-07-13
Changed list of instructions to an actual list

akamsteeg:
Minor textual tweak. :) The First Run section contains a list of instructions, but it wasn't displayed as a list.


mboelen:
Thanks for input!

PR #53, opened 2015-07-15
Fix false positive matches with unsafe ports on SQD-3624

jamesmacwhite:
The grep statement needs to be modified to prevent tagging port values that contains a value in `SQUID_DAEMON_UNSAFE_PORTS_LIST` but aren't actually equal to the listed port value


mboelen:
Reviewed and accepted as a solution. Thanks!

PR #55, opened 2015-07-24
Fix STRG-1840 (Check for disabled USB storage)

markruys:
By default, grep uses basic regular expressions (BRE). BRE needs '(', '|', and ')' escaped.


mboelen:
Thanks!

PR #56, opened 2015-07-24
Comparison error when Docker is present, but the daemon is not running (or unset).

mpepping:
When running the Container tests;

If the Docker binary is present but when the Docker daemon is not running, or DOCKER_HOST unset, the Container tests logs two error-messages, as shown below: 

```
[+] Containers
        - Docker info output (warnings)                       [ NONE ]
      - Containers
        - Total containers                                    [ UNKNOWN ]
time="2015-07-24T23:30:13+02:00" level=fatal msg="Cannot connect to the Docker daemon. Is 'docker -d' running on this host?"
        - Running containers                                  [        0 ]
./include/tests_containers: line 145: [: -gt: unary operator expected

[+] Custom Tests
  - Running custom tests...                                   [ NONE ]

```

With this pull request an extra test for the DOCKER_CONTAINERS_TOTAL variable is added (set DOCKER_CONTAINERS_TOTAL to the int 0, if string zero), which clears the integer -gt test.

Secondly, STDERR redirection to /dev/null is added for the command that sets DOCKER_CONTAINERS_RUNNING, to eliminate the "Cannot connect to the Docker daemon" error.

### DIFF
--- a/include/tests_containers
+++ b/include/tests_containers
@@ -119,6 +119,9 @@
         # Check total of containers
         logtext "Test: checking total amount of Docker containers"
         DOCKER_CONTAINERS_TOTAL=`${DOCKERBINARY} info 2> /dev/null | grep "^Containers: " | awk '{ print $2 }'`
+        if [ ${DOCKER_CONTAINERS_TOTAL} -z ]; then
+            DOCKER_CONTAINERS_TOTAL=0
+        fi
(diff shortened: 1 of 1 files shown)


mboelen:
Merged with minor change to replace -z testing. Thanks for reporting!

PR #58, opened 2015-07-27
More reliable and often faster check to determine virtualization.

markruys:
Lynis contains a check to determine whether the server is virtualized or not. On a bare metal machine, lshw can take over 10 seconds. Especially if you run `lynis update info` to check if you have the latest Lynis installed, this is undesirable. 
Much faster than `lshw` would be `dmidecode -s system-product-name`, but I'm not sure if the output could be 100% reliable be interpreted. If the system has facter installed (part of Puppet), then this will be a much more reliable and faster way to check virtualization.


mboelen:
Thanks! It has been merged with dmidecode and systemd-detect-virt routines.

PR #69, opened 2015-09-03
Do not let a hostname that is contained in "localhost" like "cal" tri…

alobodzinski:
…gger a false positive

Alexander Lobodzinski, Germany


mboelen:
Seems to be fine with -w paramter for other operating systems like Linux, AIX, *BSD. So change committed, thanks for the feedback.

PR #70, opened 2015-09-07
Wrote 'AUTH-9407' - logging failed login attempts.

kboratynski:
Initial test to login failed login attempts via login.defs configuration.


mboelen:
Thanks. It is has been merged with some minor changes and renumbered to AUTH-9408 (even numbering as much as possible).

PR #71, opened 2015-09-07
/usr/sbin/httpd on OpenBSD is builtin non-Apache webserver, do not ru…

alobodzinski:
…n with -v

If Apache is installed it is detected in /usr/local/sbin/httpd nevertheless

You can use my name in the contributions :-)


mboelen:
Committed, thanks! Please state full name + country for inclusion ;-)

PR #75, opened 2015-09-11
In quiet mode, hardening index was not written to log and report files

alobodzinski:
Calculate hardening index even in quiet mode


mboelen:
Thanks, has been merged.

PR #77, opened 2015-09-14
Bugfix for BSD: unrelated variable got logged instead of umask

alobodzinski:


alobodzinski:
Bugfix for BSD: unrelated variable got logged instead of umask

mboelen:
Merged, thanks!

PR #78, opened 2015-09-16
Freebsd services

rsmith-nl:
These changes enhance the services detection on FreeBSD.

First they add support for finding the `service(8)` binary, which is used for starting, stopping and querying services.

If available, lynis uses this binary to detect running services. If not it falls back to the original code that reads `/etc/rc.conf`.


mboelen:
Thanks, it has been merged.

PR #79, opened 2015-09-21
Bugfix: nginx SSL detection worked on Linux only (only GNU sed unders…

alobodzinski:
…tands \t on the command line)

For non-GNU sed the \t has to be replaced with a real tab character.  Code previously failed on OpenBSD and MacOS, now works there (and still on Linux).


mboelen:
Thanks, has been merged!

PR #80, opened 2015-10-04
Combine "sort | uniq" to "sort -u"

SiemKorteweg:
Combine "sort | uniq" into "sort -u" to reduce the number of processes used for running Lynis.
The busybox version of sort also supports the -u option.


mboelen:
Thanks for the commit, it has been merged.

PR #82, opened 2015-10-10
Optimized use of cat, grep and awk.

SiemKorteweg:
Less processes are created during the Lynis run.
Detection of duplicate entries in /etc/hosts is now more robust.
On CentOS7 the SERVICE_MANAGER is detected correctly.


mboelen:
Thanks, changes have been merged!

PR #83, opened 2015-10-16
Authentication fixes for OpenBSD

alobodzinski:
AUTH-9208: OpenBSD has /etc/master.passwd of course

AUTH-9234: OpenBSD real user ids run from 1000-60000 without 32767 by default; enable this test for non-Linux in the first place


alobodzinski:
Yes - ypbind gets the auth info from ypldap which basically is a NIS server with an LDAP backend.
Not really nice, but at least the sensitive, unencrypted data does not leave the localhost interface.

PR #85, opened 2015-11-04
Syslog ng

alobodzinski:
LOGG-2154: syslog-ng config file has a different syntax


alobodzinski:
Yes - ypbind gets the auth info from ypldap which basically is a NIS server with an LDAP backend.
Not really nice, but at least the sensitive, unencrypted data does not leave the localhost interface.

mboelen:
Can't merge the commit, as it has of non-related changes. Seems that it was using an outdated local version.

Proposal: clean this pull request and create several smaller ones (one piece at a time).

alobodzinski:
OK, I'll try again

PR #86, opened 2015-11-08
Add python2/3 detection and report version

jnovinger:
Add support for checking generic Python, Python 2.x, and Python 3.x with
version reporting.


mboelen:
Thanks for the pull request, it has been committed.

PR #92, opened 2015-11-27
Implemented [SSH-7426]

kboratynski:
Determines whether a reverse mapping in DNS, during SSH setting up
session, is checked.

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -154,6 +154,34 @@
 #
 #################################################################################
 #
+    # Test        : SSH-7426
+    # Description : Check SSH VerifyReverseMapping option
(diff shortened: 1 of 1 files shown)


rhyven:
Aha, this one is interesting.

Is it really worth three points?  I actually disable this (`UseDNS no`), because of the extra time it takes to connect... and because often I'm connecting from a mobile network, so I don't have reliable reverse DNS resolution.

kboratynski:
In my opinion, it is. Wrong DNS records might be a point for Man in the middle attacks.

You should always use reliable DNS server.   If you cannot, you should use your using VPN.
Would you like to get some references?

kboratynski:
Moved into: https://github.com/CISOfy/lynis/pull/100.

PR #94, opened 2015-11-27
Implemented [SSH-7618].

kboratynski:
Determines MaxAuthTries.

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -221,7 +221,32 @@
 #
 #################################################################################
 #
-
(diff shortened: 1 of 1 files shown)


mboelen:
I suggest to make it into one test if possible.
Add an additional column with the conditition (equal, not equal, less, more). Otherwise the SSH related settings will still be scattered. With a good base function, we can reuse this also for other areas.

kboratynski:
Made it easy with #105 .

PR #95, opened 2015-11-27
Implemented [SSH-7418].

kboratynski:
Determines ServerKeyBits option.


kboratynski:
No need to check as long as `ServerKeyBits` option is used for `Protocol 1` only.
It will be easy to add to #105 .

PR #100, opened 2015-12-03
features/ssh refactorization

kboratynski:
Decided to implement (modify [SSH-7408]) a simple function that combines many SSH-options.  Moved some tests into [SSH-7408] and added many others that were not implemented previously.

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -93,7 +93,8 @@
         SSHOPS="Protocol:2,,1\
                 PermitRootLogin:NO,WITHOUT-PASSWORD,YES\
                 StrictModes:YES,,NO\
-                VerifyReverseMapping:YES,,NO"
+                VerifyReverseMapping:YES,,NO\
+                IgnoreRhosts:YES,,NO"
 
         for I in ${SSHOPS};
         do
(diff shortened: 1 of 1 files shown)


rhyven:
Aha yeah that looks like a better way to do it  :)

I'll wait for @mboelen to address this, and then add a couple more goodies

kboratynski:
@rhyven, @mboelen 

Now I will try to improve [SSH-7408] with the type of test.
As you see, sometimes we need to test `=` and some `{,}`, so I will try to unify this with third parameter to `SSHOPS`.

mboelen:
Thanks for refactoring, looks like a good start.

Thought: one thing we should keep an eye on, is exceptions. Like if option A is set with value Yes, it might be a weak configuration. Unless option B is also set, and strengthen option A.

kboratynski:
@mboelen Could you be so kind and give an example of such a situation?

You mean: `PasswordsAuthentication no` and `PermitRootLogin yes` is the same as: `PermitRootLogin without-password`, am I right?

mboelen:
Yes, that would be indeed a combination where things combined can weaken or strengthen the configuration. Or with public key authentication.

PR #103, opened 2015-12-04
features/ssh configuration

kboratynski:
Added next improvements for SSH server.

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -97,6 +97,7 @@
                 IgnoreRhosts:YES,,NO\
                 UseDNS:YES,,NO\
                 X11Forwarding:NO,,YES\
+                FingerprintHash:SHA256,MD5,\
                 PrintLastLog:YES,,NO"
 
         for I in ${SSHOPS}; do
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -98,6 +98,7 @@
                 UseDNS:YES,,NO\
                 X11Forwarding:NO,,YES\
                 FingerprintHash:SHA256,MD5,\
+                TCPKeepAlive:YES,,NO\
                 PrintLastLog:YES,,NO"
 
         for I in ${SSHOPS}; do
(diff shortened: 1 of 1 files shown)


include/tests_ssh
```diff
@@ -97,6 +97,9 @@
                 IgnoreRhosts:YES,,NO\
                 UseDNS:YES,,NO\
                 X11Forwarding:NO,,YES\
+                FingerprintHash:SHA256,MD5,\
+                TCPKeepAlive:YES,,NO\
+                Compression:DELAYED,YES,NO\
```
rhyven: Re Compression:

Hmm... Compression is so often a vector for a vulnerability (e.g. BREACH, CRIME, Logjam).  I definitely wouldn't enable it before authentication, and might not enable it at all.  I'd be more inclined to put YES as the "wrong" option.

I don't know of any current vulnerabilities of SSH compression, but that doesn't mean they're not there... and the compression/decompression stack seems to be a good place to look.

Thoughts?

include/tests_ssh
```diff
@@ -97,6 +97,9 @@
                 IgnoreRhosts:YES,,NO\
                 UseDNS:YES,,NO\
                 X11Forwarding:NO,,YES\
+                FingerprintHash:SHA256,MD5,\
```
rhyven: FingerprintHash:

This one is interesting.  FingerprintHash isn't present in man sshd_config(5) under OpenSSH_6.7p1, which is current in LMDE2.  I know @mboelen is hesitant to include SSH directives which are *too* modern -- because people are likely to break their own systems by applying suggestions that aren't compatible.

Because SHA256 is the default (per http://man7.org/linux/man-pages/man5/sshd_config.5.html), is this really worth including?

include/tests_ssh
kboratynski: I was thinking of this. And on one hand have performance and on the other, vunerabillities.
I don't know them to.

How about chanigng it in reverse? I mean `Compression no` would be recomended, `delayed` would be medium scored and `yes` -- wrong one?

mboelen:
People should indeed be careful with just applying advice. One idea might be to add a minimal OpenSSH release version behind each option.

include/tests_ssh
rhyven: Looks great  :)

PR #108, opened 2015-12-14
MAIL-8838: dovecot does not have postfix as a prerequisite

alobodzinski:
and fixed some spelling


mboelen:
Thanks, changes have been merged.

PR #109, opened 2016-01-02
SSH Tests

asiebelt:
Hi,

I have made some changes to the SSH tests, because they rendered strange results for me.
I am a bit conservative in marking settings as "weak" when they are worse than the "medium" score.

Would be happy, if you could consider that patch.

Kind regards,

asiebelt

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -148,7 +148,7 @@
                     fi
 
                 elif [ "${TESTTYPE}" = "<" ]; then
-                    if [ "${FOUNDVALUE}" -ge "${WEAKVALUE}" -o "${FOUNDVALUE}" -gt "${MEDIUMSCOREDVALUE}"]; then
+                    if [ "${FOUNDVALUE}" -ge "${WEAKVALUE}" -o "${FOUNDVALUE}" -gt "${MEDIUMSCOREDVALUE}" ]; then
                         RESULT="WEAK"
                     elif [ "${FOUNDVALUE}" -le "${MEDIUMSCOREDVALUE}" ]; then
                         RESULT="MIDSCORED"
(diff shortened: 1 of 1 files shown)


mboelen:
@kboratynski can you confirm this patch is OK and has no unintended consequences on your refactoring?

kboratynski:
99fd20aae06d357cbf0a1fd854b6b954401aeb1f -- OK.
7f462d0a4136134f493ad98a5e00e9f333d4f3c0 -- OK.

@asiebelt, be so kind and explain your modification to the scoring.
The best way is to present sample input and your comment, why Lynis presented wrong results.
The commits I accepted are obvious.

asiebelt:
With the current ssh tests, I get the following results:

  - Consider hardening of SSH configuration [SSH-7408] 
    - Details: MaxAuthTries (1 --> 1)
      https://cisofy.com/controls/SSH-7408/
  - Consider hardening of SSH configuration [SSH-7408] 
    - Details: LoginGraceTime (30 --> 120)
      https://cisofy.com/controls/SSH-7408/

So, the MaxAuthTries is already optimal. Also with the LoginGraceTime I get a hardening hint, even though my configuration is better than the expected value.

We have four possible results:
(1) less than or equal to the expected value
(2) bigger than expected value and less than or equal to medium scored value
(3) bigger than medium scored value and less than or equal to weak value
(4) bigger than weak value

With my change (1) would get the result "GOOD", (2) would be "MIDSCORED" and (3)+(4) would be "WEAK".

Kind regards,

asiebelt

kboratynski:
@asiebelt -- thanks, my mistake, of course.
@mboelen -- pull these commits. ;-)

mboelen:
Thanks @asiebelt for creating the patch. It has been merged.

PR #113, opened 2016-01-16
Fix "/etc/inittab" -> "/etc/sysconfig/init"

pyllyukko:


mboelen:
Thanks, it has been merged.

PR #114, opened 2016-01-20
[NETW-3004] Support OpenBSD

alobodzinski:
[BOOT-5159] Do not throw an exception if test is skipped


mboelen:
Closing, as it is already implemented.

Issue #120, opened 2016-02-08
TCPKeepAlive no or yes?


rilindo:
I see that Lynis is flagging TCPKeepAlive if is set to no:

[17:44:59] Test: Checking TCPKeepAlive in /etc/ssh/sshd_config
[17:44:59] Result: Option TCPKeepAlive found in /etc/ssh/sshd_config
[17:44:59] Result: Option TCPKeepAlive value is  NO
[17:44:59] Result: SSH option TCPKeepAlive is in a weak configuruation state and should be fixed
[17:44:59] Suggestion: Consider hardening SSH configuration [test:SSH-7408] [details:TCPKeepAlive (NO --> YES)] [solution:-]

Why would we want to turn it on? TCPKeepAlive apparently has spoofing issues, so it is recommend that be turn off:

https://drupal.star.bnl.gov/STAR/comp/sofi/facility-access/ssh-stable-con

rhyven:
Thanks for bringing that to my attention, @rilindo.  You're quite right.  It seems that there's no actual **exploitable** vulnerability with TCPKeepAlive(1) -- yet.  But let's fix this.

After a bit of reading this morning (thanks again), I've found:

> **TCPKeepAlive**:  This option only uses TCP keepalives (as opposed to using ssh level keepalives)

vs 

> **ServerAliveInterval**:  The server alive messages are sent through the encrypted channel and therefore will not be spoofable.  The TCP keepalive option enabled by TCPKeepAlive is spoofable.

(Ref: man page for ssh_config)

It appears much more appropriate for Lynis to:
 - Ensure TCPKeepAlive is set to **No**
 - **Suggest** that ServerAliveInterval could be set to a value greater than zero (2)

@mboelen @kboratynski -- sound good?

(1) http://unix.stackexchange.com/questions/208055/is-sshd-config-tpckeepalive-still-using-unencrypted-channel-and-therefore-vulner

(2) http://unix.stackexchange.com/questions/34004/how-does-tcp-keepalive-work-in-ssh

mboelen:
Implemented. @rilindo and @rhyven: can you test if the changes give the expected on your system as well?

rilindo:
For the TCPKeepAlive, that worked. Yay! Thank you!

For ServerAliveInterval, not so much:

`- SSH option: ServerAliveInterval [ NOT FOUND ]`

And I did confirm that it is set:

[...]

PR #121, opened 2016-02-09
Removed ServerAliveInterval

rhyven:
Belongs in ssh_config instead.  Ref issue #120


mboelen:
Thanks, it is removed.

PR #134, opened 2016-03-13
Make AddHP identify non-optimal HP in logfile entries

### LINKED ISSUE #128: Feature Request - More useful output with --cronjob
I'm using a cron entry to run Lynis on my servers daily.

With `./lynis --cronjob`, the lack of formatting makes the output difficult to read in email:

> - debsums utility [ FOUND ]
> - Cron job for debsums [ FOUND ]
> - Checking security repository in sources.list file [ OK ]
> - Checking vulnerable packages [ WARNING ]

The WARNING above tends to get lost in the screed of text.

With `./lynis --cronjob -q`, the **only** output is:

> - Checking vulnerable packages [ WARNING ]

This isn't enough information for me.  In this current setup, I also get a red [ FOUND ] for a compiler, a [ NOT FOUND ] for SELinux/AppArmor/grsecurity, a [ NOT ACTIVE ] for host-based firewall... there's quite a lot that's missing.

I think there are a couple options here:
 - HTML emailing.  Probably high-effort with low returns, especially with the Lynis Enterprise Portal.
 - More granular output filtering.  Instead of just quiet mode, perhaps an option that will exclude all "[ OK ]" results.  (note I've tried to do this with grepz, but it returns strange results - see below) 
 - Perhaps the Lynis Portal could generate alerts for cronjob-executed uploads?

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

I'm trying to just use `./lynis -c -Q | grep -v '[ OK ]'`, but that doesn't respond with anything helpful:
```

root@xyz:/usr/local/lynis/lynis-master# ./lynis -c -Q | grep -v '\[ OK \]'


################################################################################

################################################################################

```

rhyven:
Helps resolve #128.

I thought doing it this way would minimise the impact on anyone who already parses the log file somehow.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -71,7 +71,11 @@
         HPADD=$1; HPADDMAX=$2
         HPPOINTS=`expr ${HPPOINTS} + ${HPADD}`
         HPTOTAL=`expr ${HPTOTAL} + ${HPADDMAX}`
-        LogText "Hardening: assigned ${HPADD} hardening points (max for this item: ${HPADDMAX}), current: ${HPPOINTS}, total: ${HPTOTAL}"
+        if [ ${HPADD} -eq ${HPADDMAX} ]; then
+            LogText "Hardening: assigned ${HPADD} hardening points (max for this item: ${HPADDMAX}), current: ${HPPOINTS}, total: ${HPTOTAL}"
+        else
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks! Item is merged, with small changes. The [*] might be confusing in the future on what it means. Instead the lines have been changed, so you can easily grep on different choice of wording.

rhyven:
Aha, perfect.  I like your change - I'd considered doing that, but thought it might break things for anyone who already had some sort of log parser running  :)

Issue #122, opened 2016-02-10
Apt-get not recognised as a package audit tool on minimal installs


rhyven:
Hi guys, not sure if this is intentional but thought I'd check.

Running Debian 8.2 here, on a minimal installation.  I get a fail on PKGS-7398.

PKGS-7398 (Check for package audit tool) relies on PKGS-7392 to set `PACKAGE_AUDIT_TOOL="apt-get"` and `PACKAGE_AUDIT_TOOL_FOUND=1`

However, 7392 won't detect apt-get if:

1. update-notifier isn't installed (Line 885), AND
2. all apt-get packages are up-to-date (Line 914)

Specifically, the command `apt-get --dry-run --show-upgraded upgrade 2> /dev/null | grep '-security' | grep "^Inst"^C` will return null if all updates are already installed.

mboelen:
With your patch it works correctly now, right?

rhyven:
Sure does!  Thanks!  :)

PR #135, opened 2016-03-13
Accept apt-get as a package audit tool

rhyven:
@mboelen what do you think of this as a first step to addressing issue #122 ?


mboelen:
Merged. Thanks!

Issue #27, opened 2015-03-09
lynis 2.0 expose_php false detection ?


centminmod:
I ran lynis 2.0 on a CentOS 7 server installed with my custom CentminMod.com LEMP stack where Nginx and PHP-FPM are source compiled.

Lynis detects and reports expose_php is ON while in fact it is OFF

        [+] PHP
    ------------------------------------
      - Checking PHP                                              [ FOUND ]
        - Checking PHP disabled functions                         [ FOUND ]
        - Checking expose_php option                              [ ON ]
        - Checking enable_dl option                               [ OFF ]
        - Checking allow_url_fopen option                         [ ON ]
        - Checking allow_url_include option                       [ OFF ]

check PHPINFO

    php -i | grep expose
    expose_php => Off => Off

PHP-FPM install has custom settings set in configscandir defined at PHP compile time

    cat /etc/centminmod/php.d/a_customphp.ini
    
    date.timezone = UTC
    max_execution_time = 60
    short_open_tag = On
    realpath_cache_size = 8192k
    realpath_cache_ttl = 600
    upload_max_filesize = 20M
    memory_limit = 160M
    post_max_size = 20M
    expose_php = Off
    mail.add_x_header = Off
    max_input_nesting_level = 128
    max_input_vars = 2000
    mysqlnd.net_cmd_buffer_size = 16384

this overrides the default php.ini set expose_php value

    grep expose_php /usr/local/lib/php.ini
    expose_php = On

so maybe extend the check to grep php -i output ?

    php -i | grep expose

centminmod:
also my additional .ini files registered by PHP configscandir compile option

    php --ini
    Configuration File (php.ini) Path: /usr/local/lib
    Loaded Configuration File:         /usr/local/lib/php.ini
    Scan for additional .ini files in: /etc/centminmod/php.d
    Additional .ini files parsed:      /etc/centminmod/php.d/a_customphp.ini,
    /etc/centminmod/php.d/curlcainfo.ini,
    /etc/centminmod/php.d/zendopcache.ini

you could add that to array of .ini files you check for php.ini which you grep for any checked php variables like expose_php and make sure those get higher priority over php.ini as they would potentially override php.ini values ?

adisatria:
how to make "expose_php = Off" in Debian ?

quantumpacket:
Correction to initial bug report, the `php.d/*` directory has nothing to do with PHP-FPM. That is what the `php-fpm.d/*` directory is for.

[...]

PR #137, opened 2016-03-21
Added ini dir of EL6

technimad:
On my EL6 systems PHP uses an ini directory of /etc/php.d, added this to the php test.


technimad:
This is somewhat related to https://github.com/CISOfy/lynis/issues/27 
I think the proposal in that thread, to use php --ini, to get the actual loaded config files might be better than maintaining a statis list in this file.

mboelen:
Thanks, merged!

PR #138, opened 2016-03-22
Hide fail2ban open+deleted file warning, when grsec is installed

rhyven:
One of the features in grsecurity also prevents Fail2Ban from creating a temp file under /var/tmp.  I've been unable to pinpoint the line in fail2ban which causes the issue, but have confirmed it's related to grsecurity.

Note: disabling the PaX MPROTECT function on /usr/bin/python will also resolve this issue, but it's better to keep MPROTECT on.


mboelen:
Thanks, merged!

Issue #118, opened 2016-01-30
Install a PAM module for password strength testing like pam_cracklib or pam_passwdqc [AUTH-9262]


azman0101:
Hi,

I'm using pam module pwquality which is enable by default in Ubuntu Wily.

But I get the suggestion AUTH-9262.

Is this normal behavior ?

mboelen:
Thanks for reporting @azman0101, that module seems to be a good alternative.

Can you show the lines used in your configuration (e.g. in your PAM itself) and the name of the pam module? Then we can add it to the test.

azman0101:
this is the line present in the file /etc/pam.d/common-password

```bash
password        requisite                       pam_pwquality.so retry=3
```

mboelen:
Hi @azman0101, can you test if recent change works for you correctly as well?

azman0101:
Hi @mboelen, Perfect ! Thxs

PR #139, opened 2016-03-22
AUTH-9262 - restructure + pwquality

rhyven:
Added pwquality (default in some Ubuntu variants) to accepted password-quality modules - addresses issue #118.

Restructured tests so that full points (3/3) are granted if any of these libraries are installed (particularly because passwdqc and cracklib are incompatible).


mboelen:
Merged, thanks!

PR #141, opened 2016-03-22
Added brew as package manager.

kboratynski:
Adds new test to detect Brew package manager and query installed packages.


mboelen:
Committed, thanks!

PR #147, opened 2016-03-25
Escape Special Character on Function "Display"

JohnTroony:
When `echo` command is used without switch `-e` which interpret escape characters, the text supposed to be colored is printed as plain text.

I've only tested this on my Arch-Linux. 
error
fixerr


mboelen:
Can't merge this request, as that will introduce an issue on other platforms. Instead, Display function replaced by echo, as part of issue https://github.com/CISOfy/lynis/issues/152

JohnTroony:
Alright, thanks.. Though I'm curious, which distribution(s) will be affected if we use `echo -e`?

mboelen:
The non-Linux ones. Think of systems like AIX, Solaris, etc.

PR #150, opened 2016-03-29
Echo used instead of cat

cnrat:
echo /etc/fstab is wrong.


rhyven:
Heh nice spotting, +1

mboelen:
Thanks for catching this one, it has been merged.

PR #154, opened 2016-04-01
Checks if DB2 LUW is running on a server

angoca:
The check if performing by doing a grep on db2sysc, the DB2 instance process.


mboelen:
Thanks for your contribution, it has been merged, and slightly optimized with the IsRunning function.

PR #156, opened 2016-04-04
Replace repetitive echo commands with single printf statement

dminca:
* replace too many `echo`s with `printf` to keep the code **DRY**
* `echo` is useful when it comes to printing the value of a variable or a (simple) line
* `printf` can basically do what the C version of it can do


mboelen:
Thanks for your suggestion, it has been merged.

dminca:
You're welcome :+1:

Issue #157, opened 2016-04-04
Prepare functions for tests


kboratynski:
To remove duplication of the code, write and implement the following functions for tests:
- `ValueEqual` (eq)
- `ValueGreaterThan` (gt)
- `ValueGreaterEqual` (ge)
- `ValueNotEqual` (ne)
- `ValueLessThan` (lt)
- `ValueLessEqual` (le)

mboelen:
Changed some names to match them with the shell equivalents (and added their short version)

mboelen:
A while ago an initial container was already created, named TestValue. Using the -- flags might greatly improve readability and forward/backward support, as this function might change. So suggesting that.

CheckValue --id "01" --description "SSH permit root login" --service "ssh" --field <etc>

mboelen:
Closing, most of this functionality is implemented. New tests will be created later when required.

PR #161, opened 2016-04-08
features/testcases

kboratynski:
Implements #157.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -68,6 +68,8 @@
 #    TestValue                  Evaluate a value in a string or key
 #    ViewCategories             Display tests categories
 #    WaitForKeypress            Wait for user to press a key to continue
+
+#    TestCase_ValueEqual        Test case for checking if value is equal to something
 #
 #################################################################################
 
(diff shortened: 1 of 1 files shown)


mboelen:
Merged, with following feedback for further improvement:

Feedback:
- Simplify testing of numbers by eliminating number of IF-statements, makes it much easier to read [1]
- Replace == with a single =
- Rename functions into the list in https://github.com/CISOfy/lynis/issues/157 (shorter, easier to read and use)
- [ "$#" -ne "2" ] can be [ $3 -ne 2 ] (does not need " signs)


[1] For example, you can rewrite the TestCase_ValueEqual test into this

# Test number of parameters
case $# in
    2)
             if [ "$1" = "$2" ]; then
                 LogText "TestCase_ValueEqual: ${1} is equal to ${2}"
                 RETVAL=0
             fi
             ;;
    3)
             if [ "$2" = "$3" ]; then
                 LogText "TestCase_ValueEqual: ${OPTION} is equal to ${3}"
                 RETVAL=1
             fi
             ;;
    *)
             <Show error: Incorrect number of arguments for function>
             ;;
esac

PR #169, opened 2016-04-20
Display skipped atomic tests

rhyven:
Added section to log & display skipped atomic tests.


mboelen:
Implemented, with a few other enhancements. We are making changes to Lynis output to only show what is relevant. So your change would actually show too much for the standard. Instead, it will now be displayed when using the --verbose option. This way it gets displayed if people want the verbose output, and skipped when running in normal mode. In all cases it will log that an atomic test is skipped:
e.g. Atomic test (SSH-7408:port) skipped by configuration (skip-test)

rhyven:
Aha, good call!

PR #173, opened 2016-04-21
Buglet prevented the --profile option from working

alobodzinski:


mboelen:
Thanks for the code. This would not solve the issue, as the profile search function has been changed.

I've made changes to allow specifying a profile. Instead of copy a default.prf to your own profile, you just have to create a custom.prf file with your changes in it. Place this custom.prf in the directory where your default.prf resides as well. Then check it with:

> lynis show profiles

mboelen:
Closing this pull, as recent changes should solve it.

Issue #179, opened 2016-04-27
Suprious warnings from PROC-3612 & PROC-3614 with grsecurity


rhyven:
When the CONFIG_GRKERNSEC_PROC kernel options are enabled, the kernel prevents access to the WCHAN mapping, therefore causing `ps -o wchan` to throw a warning message:

```
# ps -o wchan
Warning: /boot/System.map-4.3.3-grsec not parseable as a System.map
```

This function is called by PROC-3612 (line 78) and PROC-3614 (line 102), leading to spurious warnings in both normal function, as well as in the emailed output of cron jobs.

I'm not sure if the warning affects the check.  If it doesn't, I propose simply adding a `2>&1` to the end of the PSBINARY call on each of these lines.

Alternatively, we could simply remove WCHAN from the list of output parameters passed to PSBINARY.

@mboelen:
1. Do you know if the lack of wchan affects the outcome of these tests?
2. If not, do you agree with my solution?
3. Do you think we should **only** do the redirection if grsec is installed, or is naïve redirection OK?

Let me know - I'm happy to make a PR.

mboelen:
Thanks for reporting. Removed wchan from output, that doesn't do much anyways. Let's keep it simple :)

https://github.com/CISOfy/lynis/commit/84d619852aabde7c35afd7f780a16d6d91579ecc

Would be good if you could set-up a VM with a normal Debian installation (no grsecurity), with a similar configuration. Then run Lynis on both and compare the log file, to see if more tests are obstructed.

Closing this issue, as it was a quick thing to solve.

rhyven:
I didn't have the chance to stand up a separate VM, but I booted one of my test VM's into the stock Debian 3.16.0-4-amd64 kernel and ran Lynis.

I can't see any other tests being obstructed.  Ran the two logfiles through Diff and there were only very minor differences (e.g. different PIDs and temp files).  This is interesting though:

**Plain Kernel**
```
Performing test ID BOOT-5104 (Determine service manager)
Result: cmdline found = /lib/systemd/systemd
Found: systemd
```
**Grsec Kernel**
```
Performing test ID BOOT-5104 (Determine service manager)
Result: cmdline found = /sbin/init
Note: Using real readlink binary to determine symlinks
Result: readlink shows /lib/systemd/systemd as output
Result: symlink found, pointing to file /lib/systemd/systemd
Found: systemd
```

Also discovered this:

**Grsec Kernel**
```
[...]

PR #180, opened 2016-04-27
Remove wchan from PROC-3614, to finish #179

rhyven:


mboelen:
This won't work, as the awk statement after the pipe will select incorrect field. Want to resubmit?

rhyven:
Done!

PR #186, opened 2016-05-02
- [LOGG-2190]:  Running lynis from /etc/cron.daily makes some programs delete in-use files

alobodzinski:
- grep for whole words to ignore
- logging what exactly keeps a file open can be very helpful


mboelen:
Thanks, it has been merged

rhyven:
I like what you've done there, @alobodzinski :+1:

PR #188, opened 2016-05-03
Fixed GRSECFOUND and WaitForKeyPress

rhyven:
1)  WaitForKeyPress function called with incorrect capitalisation (failed)
2)  GRSECFOUND isn't set at this point in operation; reverted to checking for /dev/grsec

### DIFF
--- a/include/tests_logging
+++ b/include/tests_logging
@@ -477,7 +477,7 @@
         LogText "Test: checking deleted files that are still in use"
 
         LSOF_GREP="WARNING|Output information"
-        if [ ${GRSECFOUND} -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for the pull request.

I've changed the function call manually, and moved detection grsecurity to the OS detection part. This way it can be used on all tests and is the result stored in a new variable GRSEC_FOUND. This simplifies several tests, so seems like a better way to Rome. Let me know if my changes do not work for you. Will close this pull request (and marked it as duplicate to prevent merging).

rhyven:
That's **much** better... I was considering doing it that way too, but wasn't sure where to put the initial check.  :)

PR #191, opened 2016-05-09
Add /var/tmp to FS to check

StrangeBubble:
Expected output :

[14:31:35] File system:    /var/tmp
[14:31:35] Expected flags: nodev noexec nosuid
[14:31:35] Found flags:    defaults rw noexec nosuid nodev bind
[14:31:35] Result: GOOD, found mount option nodev on file system /var/tmp
[14:31:35] Result: GOOD, found mount option noexec on file system /var/tmp
[14:31:35] Result: GOOD, found mount option nosuid on file system /var/tmp
[14:31:35] Result: marked /var/tmp as fully hardenened

### DIFF
--- a/include/tests_filesystems
+++ b/include/tests_filesystems
@@ -479,11 +479,12 @@
     #               /home                      v              v
     #               /tmp                       v      v       v
     #               /var                                      v
+    #               /var/tmp                   v      v       v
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, this has been merged manually.

PR #192, opened 2016-05-09
FILE-6372 : error in awk comparison

StrangeBubble:
Check if mount option is part of options, instead of just checking one value.


mboelen:
Thanks, merged

PR #193, opened 2016-05-09
FILE-xyzz : Bind mount the /var/tmp directory to /tmp

StrangeBubble:
Check if a bind mount is done between /var/tmp and the directory /tmp


mboelen:
Thanks, merged.

PR #201, opened 2016-05-18
Update tests_boot_services

ratrop:
Added detection of password for Grub2 in Ubuntu 14.04 LTS. Previous version doesn't detect it.


mboelen:
Thanks for pointing this out. Adding just the FIND5 variable does not change the test. Can you show the difference in output on your system?

ratrop:
Hello,

It's true, only adjusting the FIND5 variable does not change the test. I've
send the pull request before test it. I'm sorry, is the first time that I
collaborate in a opensource project. Then I made another change, test it
and send the request again.

2016-05-19 17:30 GMT+02:00 Michael Boelen <notifications@github.com>:

Issue #197, opened 2016-05-18
SSH: Why do you recommend UseDNS?


rugk:
```
  - Consider hardening SSH configuration [SSH-7408] 
    - Details: UseDNS (NO --> YES)
      https://cisofy.com/controls/SSH-7408/
```

However such a DNS lookup [is unnecessary](https://unix.stackexchange.com/questions/56941/what-is-the-point-of-sshd-usedns-option) and may also [create a vulnerability](https://security.googleblog.com/2016/02/cve-2015-7547-glibc-getaddrinfo-stack.html).

rhyven:
Thanks @rugk.  Great question and great links.  +1, I'm convinced.

Also, I found that the OpenSSH team feels the same way; default changed to "No" in OpenSSH v6.8p1 and later: https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/424371/comments/11

I'll do a PR now for @mboelen

mboelen:
Implemented

rugk:
Nice.

@rhyven BTW you can (automatically) [let issues close](https://help.github.com/articles/closing-issues-via-commit-messages/) when a PR is merged.

rhyven:
Oh hey, that's cool -- thanks @rugk!

PR #204, opened 2016-05-18
UseDNS = No is a safer configuration

rhyven:
See Issue #197.  

References:
 - https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/424371/comments/11
 - https://unix.stackexchange.com/questions/56941/what-is-the-point-of-sshd-usedns-option
 - https://security.googleblog.com/2016/02/cve-2015-7547-glibc-getaddrinfo-stack.html


mboelen:
Thanks, makes sense.

PR #205, opened 2016-05-19
Remove the support of some filesystems

StrangeBubble:
[removed links]

Some chapters ask to remove the support of some filesystems : cramfs hfs hfsplus squashfs udf freevxfs jffs2

Tested on Red Hat 7.2 and CentOS 7.2

### DIFF
--- a/include/tests_filesystems
+++ b/include/tests_filesystems
@@ -644,6 +644,43 @@
 #
 #################################################################################
 #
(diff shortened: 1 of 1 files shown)


mboelen:
Looks like a good addition. What if we would change this slightly: instead of putting the file systems in the test itself, we move them to the default profile.

PR #208, opened 2016-05-30
improve non-interactive use of zypper

andreasstieger:
Improve non-interactive behavior of zypper.

* Always run with -n (non-interactive) to prevent questions that block execution of lynis
* for PKGS-7328 only list packages


mboelen:
Thanks, great addition. It has been merged.

PR #210, opened 2016-06-01
Filter out open+deleted files held by MySQL <5.6

### LINKED ISSUE #209: LOGG-2190 always throws a warning with MySQL <5.6 installed
MySQL 5.5 and previous versions have a bug that leaves dozens of deleted-but-held files in /tmp.

It's been recognised since 2005 (ref http://bugs.mysql.com/bug.php?id=15089), and has been fixed in 5.6+ (ref http://bugs.mysql.com/bug.php?id=66237), but hasn't been fixed in 5.5 and it looks like it won't be.

What do you guys think about adding an exclusion for mysql files in LOGG-2190?

rhyven:
Resolves #209

### DIFF
--- a/include/tests_logging
+++ b/include/tests_logging
@@ -477,6 +477,14 @@
         LogText "Test: checking deleted files that are still in use"
 
         LSOF_GREP="WARNING|Output information"
+        
+        EARLY_MYSQL=`dpkg -l | egrep mysql-server-5.[0-5]`
+        
+        if [ ! "${EARLY_MYSQL}" = "" ]; then
+            # MySQL versions prior to 5.6 leave lots of deleted in-use files in /tmp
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for the patch. It needs a little bit more work, as this would only work on systems using the dpkg package managers. Probably a better check: use which mysqld 2> /dev/null to determine if we have a MySQL daemon. Then query its version (-V?) and strip out all other lines.

Will you try to patch it further? Let me know if you need guidance.

mboelen:
@rhyven do you have time to alter the patch slightly to make it work on all platforms? Let me know if you can't (e.g. due to time constraints).

rhyven:
@mboelen sorry mate I don't think I'll have time to do this  :-(

mboelen:
Merged, we will make some adjustments for the test.

PR #224, opened 2016-07-10
expect value of sysctl:kernel.kptr_restrict to be 2

lpirl:
from https://lwn.net/Articles/420403/:

>   The %pK format specifier is designed to hide exposed kernel
>   pointers, specifically via /proc interfaces.  Exposing these
>   pointers provides an easy target for kernel write vulnerabilities,
>   since they reveal the locations of writable structures containing
>   easily triggerable function pointers.  The behavior of %pK depends
>   on the kptr_restrict sysctl. […] If kptr_restrict is set to 2,
>   kernel pointers using %pK are printed as 0's regardless of
>   privileges.


mboelen:
Thanks, accepted and merged.

Background: specific key is available since 2010, which should be enough time now for most Linux distros to have a kernel after that period.

PR #228, opened 2016-07-14
fix: Test names are different

noone-silentrunner:
After changes in 9aa5736 the test names are different. This results in something like BOOT-5122 = boot-5122 and skipped test aren't skipped anymore. Changed from lower to upper, to match display commit.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -1793,13 +1793,13 @@
 
         # Skip test if it's configured in profile (old style)
         if [ ${SKIPTEST} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
How do you currently have the tests in your configuration file? Can you show an example?

noone-silentrunner:
In the same way described in default.prf

https://github.com/CISOfy/lynis/blob/master/default.prf#L61

Usually i do something like this:
```bash
# GRUB Password
skip-test=BOOT-5122
# Full Disk /home /tmp /var
skip-test=FILE-6310
```

I copy the string from lynis.log and paste it in my custom.prf

mboelen:
Thanks, that makes sense. Merged.

PR #237, opened 2016-07-24
i18n: wrong indentation of result column

atao60:
Display function uses wc to calc the number of spaces between label and result; wc requires LANG value to deal with multi-bytes characters, e.g. accentuated ones in french with utf-8.
But atm LANG is unset inside include/consts. Then 'wc -m' returns the number of bytes, not the number of unicode code points, i.e. "characters". On the concerned raws, the result is misaligned (too closed of the label).

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -508,8 +508,10 @@
             fi
 
             if [ ${SHOW} -eq 1 ]; then
-                # Display (counting with -m instead of -c, to support language locale)
-                LINESIZE=`echo "${TEXT}" | wc -m | tr -d ' '`
+                # Display:
+                # - counting with -m instead of -c, to support language locale
+                # - wc needs LANG to deal with multi-bytes characters but LANG has been unset in include/consts...
+                LINESIZE=`export LC_ALL= ; export LANG="${DISPLAY_LANG}";echo "${TEXT}" | wc -m | tr -d ' '`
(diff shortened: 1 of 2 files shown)


mboelen:
Can you show the output of something that goes wrong? Right now the patch would actually break the primary reason why LANG is unset in const (some commands would break on non-English output).

atao60:
> Can you show the output of something that goes wrong?

See the attached file, section "Software: Malware". There are thee lines with one accentuated character. Each one requires two bytes then the misaligned result column. The misalignment increases with the number of multi-bytes characters inside the label.

> Right now the patch would actually break the primary reason why LANG is unset in const (some commands would break on non-English output).

No. The inset LANG stays unset. 

It's possible to check up on this with a 'locale' command just after the instruction "LINESIZE=`export LC_ALL= ; export LANG="${DISPLAY_LANG}";echo[...]`" inside the Display function.

The DISPLAY_LANG variable keeps the original value of LANG. It is used only inside the above instruction. The scope of 'export LANG' is limited to the context of the command substitution defined by the backticks.


[lynis-pr237.log.txt](https://github.com/CISOfy/lynis/files/380315/lynis-pr237.log.txt)

mboelen:
You are right. Thanks, merged.

PR #242, opened 2016-07-28
Fix description of CheckFilePermissions

quant67:
"Adds a system to a group, which can be used for categorizing" should belong to 
AddSystemGroup but not CheckFilePermissions


mboelen:
Thanks for this pull request!

It looks like you just patched an outdated file and now includes a few more lines than it needed. Can you edit your patch? This way you get the contribution :)

PR #243, opened 2016-07-29
Fix description of CheckFilePermissions

quant67:
"Adds a system to a group, which can be used for categorizing" should belong to 
AddSystemGroup but not CheckFilePermissions


mboelen:
Thanks, changes merged.

Issue #245, opened 2016-07-29
bug in yum gpgcheck detection


beelsr:
this snippet, while valid and `working,` is not recognized by lynis:
`gpgcheck          = 1
`

and here's a diff to fix. 

```
diff -U0 tests_ports_packages~ tests_ports_packages
--- tests_ports_packages~       2016-07-14 14:19:10.000000000 -0400
+++ tests_ports_packages        2016-07-29 12:50:52.647929289 -0400
@@ -826,2 +826,2 @@
-           SearchItem "^gpgenabled=1$" "/etc/yum.conf"; if [ ${ITEM_FOUND} -eq 1 ]; then FOUND=1; fi
-           SearchItem "^gpgcheck=1$" "/etc/yum.conf"; if [ ${ITEM_FOUND} -eq 1 ]; then FOUND=1; fi
+           SearchItem "^gpgenabled\s*=\s*1$" "/etc/yum.conf"; if [ ${ITEM_FOUND} -eq 1 ]; then FOUND=1; fi
+           SearchItem   "^gpgcheck\s*=\s*1$" "/etc/yum.conf"; if [ ${ITEM_FOUND} -eq 1 ]; then FOUND=1; fi
```

kboratynski:
@beelsr Thank you for this report.
Would you be so kind and write a pull request?
We would appreciate your contribution.

_Postscriptum_: I am not fairly sure, yum-repo-file accepts only *one* space between parameter and `=` mark. Would you be so kind and check it? If so, I would suggest to modify your regexp to handle *only* proper yum-repos. What do you think?

beelsr:
so, I deleted my MariaDB key from my rpmdb via `rpm -e gpg-pubkey-1bb943db-511147a9`.

I then:
edited yum.conf to say `gpgcheck = 1`
verified no gpgcheck entry existed in `yum.repos.d/mariadb.repo`
issued a `yum  --disablerepo="*" --enablerepo="mariadb" install mariadb-test` command and was rewarded with the output of:
```
Downloading packages:
warning: /var/cache/yum/x86_64/7/mariadb/packages/MariaDB-10.1.16-centos7-x86_64-test.rpm: Header V4 DSA/SHA1 Signature, key ID 1bb943db: NOKEY
Retrieving key from https://yum.mariadb.org/RPM-GPG-KEY-MariaDB
Importing GPG key 0x1BB943DB:
[...]

PR #247, opened 2016-07-30
fix yum gpgenabled check to allow spaces around =

beelsr:
Issue #245 

yum.conf allows an arbitrary number of spaces to surround the equals assignment.


mboelen:
Thanks, merged.

PR #251, opened 2016-08-03
Simple [ -z {$VAR} catches unset variables only in bash.  Need to add…

alobodzinski:
… +word to correctly work on non-bash shells.

Just seeing that a lot of -z tests have crept in - you may want to look at those as well...

### DIFF
--- a/lynis
+++ b/lynis
@@ -503,8 +503,8 @@ ${NORMAL}
     Report "os_fullname=${OS_FULLNAME}"
     Report "os_version=${OS_VERSION}"
     if [ "${OS}" = "Linux" ]; then Report "linux_version=${LINUX_VERSION}"; fi
-    if [ ! -z "${OS_KERNELVERSION}" ]; then Report "os_kernel_version=${OS_KERNELVERSION}"; fi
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for the feedback.

We added them on purpose to make the code shorter. Another possible benefit could be speed, as no string comparison has to be done. I don't think -z will do checking for unset variables. It only checks if the value is not empty. For example, AIX knows -z as well: https://www.ibm.com/support/knowledgecenter/ssw_aix_61/com.ibm.aix.cmds5/test.htm

You received a variable is unset message with the GitHub version? That is a recently enabled test for the development version, to catch them and properly initialize all variables. I suggest to close this pull request, and instead have the variables initialized in include/consts. Agree?

alobodzinski:
Well, I think relying on variable is unset messages only makes real sense if you have 100% test coverage which few people have, and {$VAR+x} is more defensive...  But I can close the pull request and put the vars into include/consts - hope not too many more will crop up :-)

mboelen:
I would prefer to do both if many more show up. Strict mode (where unset variables are checked) only applies to the first phase of Lynis (before tests), and only when it is a dev version (which is GitHub version). So I don't think many will show anymore, as most were already set.

Background information: This addition is exactly to introduce a more defensive programming style and make sure that the right variables are used (and catch any typos).

PR #255, opened 2016-08-09
Hu translation

paldiz:
Add Hungarian language support for Lynis.


mboelen:
Thanks for the translation.

I see some additional commits in it, which seems not to be relevant. It looks like they are old patches? I will merge things and see if any changes are needed.

PR #258, opened 2016-08-11
Make IsVirtualMachine detect centos6.8 on openstack

quant67:
These function made mistake in centos 6.8 virtual machine. Only dmicecode work
on these environment and it return Openstack Nova. A openstack case is needed
in codes that check $SHORT codes.


mboelen:
Thanks, merged.

rhyven:
This totally wins the award for fastest merge ever!  +1

PR #263, opened 2016-08-18
Fix FILE-6344 bug

cnrat:
The result of mount command returns like
/proc on /proc type proc (rw,hidepid=2)

Witch means the regular expression should be '^/proc' not '^proc' or you can grep nothing. :P


mboelen:
Thanks, it has been merged.

PR #265, opened 2016-08-26
Fixed cut-and-paste error

alobodzinski:
This made Lynis' last message show up even in --cronjob -q mode (instead of the tooltip).  Pretty confusing...


mboelen:
Thanks, merged!

PR #268, opened 2016-08-29
Ignore GlusterFS for nodev and nosuid checks

markruys:
GlusterFS mounts use fuse and are nodev and nosuid by default. Also nodev and nosuid are invalid mount options. So ignore GlusterFS mount points in check FILE-6374.


mboelen:
Thanks, merged!

PR #278, opened 2016-09-08
Fix sed syntax for ServerAlias removal

DjSlash:
When a Apache config directive is in use, it has whitespace(s) or nothing at all prepended. Assuming that it always has a space before it doesn't have to match.

I thought I'd submitted this PR already, but it seemed not to have happened, strangely enough ...


mboelen:
Merged, thanks!

PR #281, opened 2016-09-17
Skip NTP checks from within a FreeBSD jail

inferiorhumanorgans:
As jails get their time from the host and do not have their own clocks, it's not useful to run a time synchronization program from within the jail.


mboelen:
Thanks, it has been merged.

PR #284, opened 2016-09-23
Add support for Rudder configuration management

FlorianHeigl:
Rudder (http://www.rudder-project.org) is a config management tool with automatic reporting / dynamic policy adjustment.
The "dynamic" bit happens using FusionInventory.
The config management part uses CFEngine under the hood.
I don't know if hw/os inventory tools also matter in the Lynis report, if yes, I can look into that too.

This patch extends the path searched to detect a running CFEngine agent (and fileserver daemon)
Since the agent is the same, this should immediately detect it.


ncharles:
+1 ! Thank you @FlorianHeigl

mboelen:
Thanks, it has been merged.

PR #293, opened 2016-10-18
[MALW-3280] Bitdefender AV Support (macOS)

marcus-cr:
Added support to detect if Bitdefender Antivirus for Mac is running.


mboelen:
Thanks, merged

Issue #264, opened 2016-08-21
machine-role=desktop - Options Seem Extreme For Desktop...


Geyup:
I have my .prf for `machine-role=desktop`, using a stand alone, single user, at home only desktop, I find Lynis options a bit extreme.

Don't take me wrong, I love lynis and thank you very much, but for a desktop user this needs to be toned down, you're going to scare the Newbies! LOL...

I understand trying to find a happy balance here for a desktop user a possible bit of a task, but I think the desktop feedback when running Lynis could stand for some triming back.

Maybe in time there can be something created as a more easier `desktop` option, seems a little to Fort Knox as the Americans would say for a desktop. 

My Suggestions;

**. Bootloader - Password option presence should be a `Suggestion`

**. Linux single user mode authentication - Starting or Logging in? I'm not sure in Slackware you can get around it and log in without the root password...

**. umask for a home user - `weak` in /etc/profile is relevant based on needs, and a home user doesn't need to consider changing. I've used umask 022 for many years without problem as a desktop user... Should be changed to `Suggestion` for `machine-role=desktop`, servers as weak is a more appropriate label...

**. Testing swap partitions - label should be changed to `Not Found`, swap isn't needed with large amounts of ram, I stopped using swap YEARS ago. :)

**. Testing proc mount - if this is about hiding it from a user, then in Slackware as an example it doesn't work loading it from fstab, it needs to be added to rc.local like this;

mount -o remount,hidepid=2 /proc

So testing for /proc in fstab for an entry doesn't work on all distros...

**.  Checking usb-storage driver - extreme for a single user at home desktop user. For `machine-role=desktop` this should be changed to `Suggestion`. Same thing for firewire...

**. /etc/issue contents listed as weak, should be changed to `Suggestions` for `machine-role=desktop`

**.  /etc/issue.net contents listed as weak, should be changed to `Suggestions` for `machine-role=desktop`

**.  Installed compiler(s) - compiler hardending on a home box is extreme, should be `Suggestion`

**. Installed malware scanner - should be `Suggestion` when not found, not red not found and then an X below, because I've never run a malware scanner ever as a home desktop user in Linux. 
It depends on the needs, Red wording and a X not needed. `Suggestion` and `Yellow` question ? mark, better for `machine-role=desktop`

I hope in time we will see a better based output geared towards the Desktop user, it still looks to much like Server output IMHO...

mboelen:
[...]

PR #295, opened 2016-10-20
machine-roles_issue-264

marcus-cr:
Fixed issues with #264. Tests to be amended if still needed:
BOOT-5139
AUTH-9328
FILE-6344


mboelen:
This doesn't merge, as the code was applied on an out-of-date tree. Can you resubmit?
(BTW no need to change the changelog, we can do that while doing review work and merging)

marcus-cr:
Hmm, I think I'll have to close the pull request and launch a new one? I can only comment.

mboelen:
Yes, close it and open a new one please. This is because it is in conflict. Sadly enough, it is not possible to do an inline edit quickly and resolve it.

PR #297, opened 2016-10-20
Issue264 updates

marcus-cr:
Fixed tests and profile for issue #295


mboelen:
See remark of https://github.com/CISOfy/lynis/pull/298

PR #301, opened 2016-10-24
[HRDN-7230] Result colors changed, added result log text

marcus-cr:
Results for malware scanner will show  different color for “personal”
machine-roles (in yellow) rather than the other roles (in red).

Added for leniency towards typical personal PC users for not having
malware scanner (should be a suggestion).

Also included log text result if no scanner found, regardless of
machine-role.


mboelen:
Thanks, merged!

PR #302, opened 2016-10-24
[BOOT-5139] LILO suggestions changed for new machine-roles

marcus-cr:
Test will display result in red if no boot loader password set and if
server or workstation role was set. Cosmetic change for log text, one
instead of two log texts.

Test displays result in yellow if using personal machine role as
suggestion for typical users.


mboelen:
Merged and changed. Thanks!

PR #303, opened 2016-10-24
[FINT-4315] Warning for missing AIDE conf file

marcus-cr:
Warning given if AIDE is detected but no configuration file found.
Without configuration file AIDE will not work.


mboelen:
Merged, thanks!

PR #304, opened 2016-10-24
[FILE-6332] Skips test FILE-6336 if no swap detected

marcus-cr:
Lynis will skip test FILE-6336 (check swap mount options) if no swap
partitions are found in /etc/fstab from previous test FILE-6332.

### DIFF
--- a/include/tests_filesystems
+++ b/include/tests_filesystems
@@ -261,6 +261,7 @@
           else
             Display --indent 2 --text "- Query swap partitions (fstab)" --result "${STATUS_NONE}" --color YELLOW
             LogText "Result: no swap partitions found in /etc/fstab"
+            skip-test=FILE-6336
         fi
     fi
 #


marcus-cr:
`skip-test=` only works in the profile configuration then, correct? 

The goal is to skip the swap mount options (FILE-6336) if swap wasn't detected in the previous test (FILE-6332). I noticed this while testing for #264.

mboelen:
Yes, that only works in profiles. Might be better to use a PREQS_MET type of test to skip a full test.

marcus-cr:
OK, I'll look into that to improve this test. Thanks for the guidance, as well as the information you emailed me regarding macOS file permissions.

mboelen:
Let's close this pull request then?

marcus-cr:
Sure, closing pull request until a PREQS_MET type of test can be implemented.

PR #308, opened 2016-10-27
Updates to include/tests_authentication

jpartain89:
The LogText Output for test 9208 was worded oddly. 
"Non unique UIDs can riskful for..." has been changed to:
"Non unique UIDs can be a risk for..."

- and - 

Test 9234, specifically finding the Real User info, originally using `/etc/passwd`.  macOS does NOT use `/etc/passwd` to store the user info, but rather the default system users. macOS uses `opendirectoryd` for the backend user service. Thus, the program `dscacheutil` is whats recommended to query the various items in opendirectoryd. I left comments in the test section with the basic commands you can use as reference for building out the correct FIND commands.


mboelen:
Thanks, merged!

PR #310, opened 2016-10-31
Fix for NAME-4408: getent hosts localhost   may show two lines on Ope…

alobodzinski:
…nBSD


mboelen:
Looking good, merged. Agree the second comparison is better :) Thanks!

PR #318, opened 2016-11-07
Fixes for directories with spaces in their names - macOS.

novaksam:


novaksam:
Give me a few minutes; I found some more.

novaksam:
Actually, this looks like it's more trouble than it's worth, especially since I don't know how this will affect other operating systems. I'll close it

PR #323, opened 2016-11-10
Update README.md and Chinese language support

apemost:
* https://github.com/CISOfy/lynis.git is better for git clone
* Delete trailing whitespace
* Add `GEN_LATEST_VERSION` in Chinese language


mboelen:
Merged. Thanks!

PR #328, opened 2016-11-13
Translate EN for PT

Dayvidson:
Translation to Portuguese


mboelen:
Thanks, it has been merged!

PR #342, opened 2016-12-28
Variable name in reportWarning function

angoca:
The variable message was incorrectly written


mboelen:
Thanks for the discovery and fix :)

PR #346, opened 2017-01-04
Update functions

Slydder:
Fix for apache config file names with special characters.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -715,7 +715,7 @@
         CANREAD=0
         RETVAL=1
         LogText "Test: check if we can access ${sFILE}"
-        escaped_sFILE=$(echo $"sFILE" | sed "s/\*/\\*/; s/\?/\\?/")
+        escaped_sFILE=$(echo "$sFILE" | sed "s/\*/\\*/; s/\?/\\?/")
 
         # Check for symlink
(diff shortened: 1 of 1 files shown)


quantumpacket:
@Slydder Shouldn't those be double quotes? Otherwise you are having it check the literal string `${sFILE}` and not the value of that variable. https://www.gnu.org/software/bash/manual/html_node/Quoting.html#Quoting

Slydder:
attached you will find a working version. changes include:

1. I added a new variable to escape '*' and '?' characters in the file name
2. changed the '-L' tests in the if statements (which doesn't work with escaped characters for some reason) to '-h' which seems to work
3. changed the ls -d -l calls in the if statements to stat -c %A which work with escaped characters

[functions.zip](https://github.com/CISOfy/lynis/files/689775/functions.zip)

mboelen:
Thanks for the pull request, very much appreciated!

The stat command is not available on all platforms (like AIX). So currently this patch would break core functionality and needs a little bit more rework to get accepted.

mboelen:
Manually picked some of the changes from this pull request. As said, the stat command is not working the same on all platforms. Thanks for your contribution!

PR #347, opened 2017-01-05
Fix for 3 issues

noci2012:


mboelen:
Thanks for these improvements. Unfortunately they can't be merged in the current form. This is because of several areas, like file globbing which is not working in bourne shell (/bin/sh). Can you split the changes?

Happy to discuss them. Contact me by email with a reference to this pull request.

mboelen:
Most items are fixed in the meantime with external changes. Closing this pull request.

PR #348, opened 2017-01-07
Set MAC_FRAMEWORK_ACTIVE=1 if grsecurity's RBAC is enabled

pyllyukko:
This is looked up with the ```gradm``` tool.


mboelen:
Thanks, it has been merged!

PR #357, opened 2017-02-20
MAIL-8818: $LINUX_VERSION is not set on every OS

alobodzinski:
Check if Linux version is set before using it in Postfix test


mboelen:
Thanks, merged!

PR #360, opened 2017-03-01
Add /usr/local/etc/zend/php.ini as possible php path

sigio:
Add a path to check for php.ini file

### DIFF
--- a/include/tests_php
+++ b/include/tests_php
@@ -36,6 +36,7 @@
                 /private/etc/php.ini \
                 /var/www/conf/php.ini \
                 /usr/local/etc/php.ini /usr/local/lib/php.ini \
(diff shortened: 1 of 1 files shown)


noci2012:
[+] PHP
  - Checking PHP                                              [ NOT FOUND ]

While:
19289 ?        Ss     0:00 php-fpm: master process (/etc/php/fpm-php5.6/php-fpm.conf)
19290 ?        S      0:00 php-fpm: pool www
19291 ?        S      0:00 php-fpm: pool www
19292 ?        S      0:00 php-fpm: pool www
19293 ?        S      0:00 php-fpm: pool www
19294 ?        S      0:00 php-fpm: pool www

So here 's some failed checking

PR #364, opened 2017-03-06
umask can be 0027 or 0077

saez0pub:
Actually : 
```
Test: Checking umask value in /etc/init.d/functions
Result: found umask 0027, which could be improved
```

Expected : 
```
Test: Checking umask value in /etc/init.d/functions
Result: umask 0027 is considered a properly hardened value
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Changes:
 
 Tests:
 ------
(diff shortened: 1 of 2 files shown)


mboelen:
Thanks, great feedback and nice pull request. It has been merged!

PR #366, opened 2017-03-07
Lots of cleanups

hlein:
Documentation fix for the SafePerms function.

And then, a zillion whitespace fixes for indentation in if/fi blocks; see the specific commit message.

Then a couple of small typo/syntax error fixes.


mboelen:
Thanks for another pull request.

Issue #214, opened 2016-06-12
lynis not working on solaris 10


andreas63741:
On Solaris 10 (SunOS sol10 5.10 Generic_147148-26 i86pc i386 i86pc), some parts of lynis are not working, because certain commands or command options are not supported.

I started lynis in user mode (no admin rights).
- "id -u" option not supported, "/usr/xpg4/bin/id" may be not installed
- "date +%s" option not supported
- "find -maxdepth" option not supported
- "awk" errors because of usage of unsupported options
- INTERFACES_TO_TEST in include/functions is too static

See some fixes and all errors below 

Fixes:

```
sol10(user):~/lynis -> diff lynis.x lynis
86,87d85
<     SOLARIS="0"
<
90,94c88,91
<     if [ `uname` = "SunOS" ]; then
<       SOLARIS="1"
<       MYID=`id | tr '=' ' ' | tr '(' ' ' | awk '{ print $2 }' 2> /dev/null`
<     else
<       MYID=`id -u 2> /dev/null`

---
>     if [ -x /usr/xpg4/bin/id ]; then
>         MYID=`/usr/xpg4/bin/id -u 2> /dev/null`
>       else
>         MYID=`id -u 2> /dev/null`
642c639
<     if [ "$SOLARIS" = "0" ] && [ ${NOW} -gt ${RELEASE_PLUS_TIMEDIFF} ]; then

---
>     if [ ${NOW} -gt ${RELEASE_PLUS_TIMEDIFF} ]; then

Improvement for INTERFACES_TO_TEST:

sol10(user):~/lynis/include -> diff functions.x functions
586c586
<                     INTERFACES_TO_TEST=`netstat -i | grep -v ^Name | awk '{print $1}'`

---
>                     INTERFACES_TO_TEST="e1000g1 net0"

----

sol10(user):~ ->  bash lynis_start.sh
ifconfig: status: SIOCGLIFFLAGS: net0: no such interface
cat: cannot open /etc/ssh/ssh_host_dsa_key
./include/tests_authentication: line 436: /usr/bin/logins: Permission denied
find: bad option -maxdepth
find: [-H | -L] path-list predicate-list
find: bad option -maxdepth
find: [-H | -L] path-list predicate-list
awk: syntax error near line 1
awk: bailing out near line 1
grep: can't open /etc/cups/cupsd.conf
grep: can't open /etc/cups/cupsd.conf
open device: Permission denied
awk: syntax error near line 1
awk: bailing out near line 1
[...]

PR #367, opened 2017-03-08
A bunch of Solaris compatibility tweaks

hlein:
I found that I still had a few problems in or like those listed in https://github.com/CISOfy/lynis/issues/214

With these changes, lynis runs without error for me on Solaris 10.

There may still be a few things that are sub-optimal on Solaris.


hlein:
Oh, I see!  Reverted the change of the HOSTID part.  Note that this will error if openssl is so old it does not even support sha1.  ...But I don't know if such a thing exists.

mboelen:
Thanks for the changes. Yes, that is an acceptable risk considering the age of SHA1. In the future, we might slowly migrate to some other hashing of the host identifier and then it could be a good method to use yours.

Really appreciate your quality work here. Thanks again :)

hlein:
You're welcome!  Thanks for the quick feedback and merges.

PR #369, opened 2017-03-12
Added new php paths

noci2012:
removed php5.4,  added 7.0 and 7.1 and the flavours that are known on gentoo
(apache2, cgi, cli, embed,  fpm)


mboelen:
Thanks, merged

PR #370, opened 2017-03-17
Fixed detection of security repositories

hurta2yaisel:
We have some security repositories URLs like http://debian.uci.cu/security that are not recognized by Lynis.


strangelittlemonkey:
Looks simple enough to me.

mboelen:
Thanks, it has been merged

PR #371, opened 2017-03-22
Fix regex to disregard locked accounts

### LINKED ISSUE #362: AUTH-9288 can have false positive when password is locked
Hello.

I'm experiencing a FP with AUTH-9288 with an account that has it's password locked.

The thing is, that (for some reason), the particular system account has an entry in ```Password expires``` field and the regex does not match ```:!!:```.

So the [tests_authentication](https://github.com/CISOfy/lynis/blob/master/include/tests_authentication) has:
```
# Skip fields with a !, *, or x, or !* (field $3 is last changed, $5 is maximum changed)
FIND=$(${EGREPBINARY} -v ":[\!\*x](\*)?:" /etc/shadow | ${AWKBINARY} -v today=${DAYS_SINCE_EPOCH} -F: '{ if (($5!="") && (today>$3+$5)) { print $1 }}')
```
I guess the regex could be something like ```":[\!\*x]([\*\!])?:"``` to match also ```:!!:```? Then again, maybe it should be written even more specific, so it wouldn't match e.g. ```x*```. Something like ```":(x|\*|\!(\*|\!)?):"```, which would match **exactly** ```x```, ```*```, ```!```, ```!*``` and ```!!```?

Also the test matches locked accounts with password hash, e.g. ```:!!$6...```. I think if the password is locked, it shouldn't be a problem.

pyllyukko:
This way, accounts that have ":!!:" in shadow and have an entry in
"Password expires" field don't get flagged with "Result: password of
user XYZ has been expired" by AUTH-9288.

Fixes #362

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -825,7 +825,7 @@
             LogText "Data: Days since epoch is ${DAYS_SINCE_EPOCH}"
             LogText "Test: collecting accounts which have an expired password (last day changed + maximum change time)"
             # Skip fields with a !, *, or x, or !* (field $3 is last changed, $5 is maximum changed)
-            FIND=$(${EGREPBINARY} -v ":[\!\*x](\*)?:" /etc/shadow | ${AWKBINARY} -v today=${DAYS_SINCE_EPOCH} -F: '{ if (($5!="") && (today>$3+$5)) { print $1 }}')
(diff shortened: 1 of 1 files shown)


pyllyukko:
I'm still thinking whether this test should disregard locked accounts completely (e.g. ```:!!$6...```).

mboelen:
Thanks, it has been merged. For now, let's leave the test like this. This way these accounts at least show up and the admin can take action.

PR #372, opened 2017-03-23
Added quote marks on SSH commands

guyav:
Added quote marks for the SSH commands in steps and 6.


mboelen:
Thanks for your contribution, it has been merged.

PR #377, opened 2017-04-15
Add Additional php.ini Paths

quantumpacket:
It wasn't finding my PHP config located at `/etc/php/php.ini`, added that path and some variations.


mboelen:
Thanks, it has been merged!

PR #389, opened 2017-05-02
Adds Protected Links Checks - Closes #386

### LINKED ISSUE #386: Add Protected Links Checks
Add support for hardlink and symlink file protections per kernel filesystem docs:
https://www.kernel.org/doc/Documentation/sysctl/fs.txt

Safe:
> fs.protected_hardlinks=1
fs.protected_symlinks=1

Unsafe:
> fs.protected_hardlinks=0
fs.protected_symlinks=0

quantumpacket:
Closes #386 
Add support for hardlink and symlink file protections.

### DIFF
--- a/default.prf
+++ b/default.prf
@@ -178,6 +178,9 @@ config-data=sysctl;security.bsd.unprivileged_proc_debug;0;1;Unprivileged process
 config-data=sysctl;security.bsd.unprivileged_read_msgbuf;0;1;Unprivileged processes can not read the kernel message buffer;sysctl -a;-;category:security;
 
 # Kernel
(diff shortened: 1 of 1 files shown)


mboelen:
Merged, thanks!

quantumpacket:
@mboelen I had added the line for `fs.suid_dumpable` as you suggested, but since it has multiple safer options it will mark points against you if it's a stricter value than `1`, like `2` for example. I posted in #389 about how it's being checked in `tests_kernel` and how `kernel.suid_dumpable` is not being checked for the other values. We might want to revert this commit, or remove the `fs.suid_dumpable` line, and also make sure the older kernel option is being checked for other values if it supports it.

PR #393, opened 2017-05-12
250 fixes

mslifcak:
Hi,
I compared 2.4.0 with 2.5.0, found a few problems, and offer these fixes in compliance with the license restrictions provided for this software project.
1. include/functions - the use of HasData on line1462 is not correct. Revert the change.
2. A list of locations for php.ini has one missing ROOTDIR prefix.  Added.
3. Search for certificates was expanded in a way that favored newer version of the 'find' command.  This was re-factored so that older systems can also benefit from the expanded search.

Let me know if there is further explanation needed.
Kind regards,
-Mike Slifcak

### DIFF
--- a/include/tests_crypto
+++ b/include/tests_crypto
@@ -44,7 +44,7 @@
                 if [ ${CANREAD} -eq 1 ]; then
                     LogText "Result: found directory ${DIR}"
                     # Search for CRT files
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for the great eye for detail and enhancements. They have been merged.

PR #398, opened 2017-05-26
A250 2

mslifcak:
sort list of startup scripts before processing
    sort list of certificates before processing
    sort list of services before processing
    fix missing ROOTDIR prefix


mboelen:
Thanks, another batch of good improvements. Appreciate your pull requests and attention for detail :+1:

PR #399, opened 2017-05-27
Add a check for csf testing mode

mostafahussein:
This should notify the one who make audits that CSF is in testing mode as sometimes you may forget to  disable it on a production environment which leads to flushing the firewall every N seconds. What do you think ?


mboelen:
Thanks for your pull request. I've reviewed it and merged it.

PR #401, opened 2017-06-08
add suhosin status test

mostafahussein:
As you check for disabled functions within suhosin configuration I think this kind of test might be useful. I couldn't think in a better way in order to achieve it. what do you think ?


mboelen:
Thanks, it has been merged

PR #403, opened 2017-06-09
Fix pkg audit option

jlduran:
-F, --fetch
           Fetch the database before checking.

[ci skip]


mboelen:
Thanks, it has been merged

PR #404, opened 2017-06-11
Added kernel.dmesg_restrict to sysctl checks.

dvehrs:
Preventing an unprivileged user from viewing the kernel log buffer may be an important part of general hardening of the system.   This can help prevent an attacker from gathering debugging information that might be valuable during exploitation.

It's a simple 1 line addition to default.prf, what do you think?


mboelen:
Thanks, it has been merged

efkaxlog:
:::::uuuuu?y?egygtgcy

PR #412, opened 2017-07-03
Update tests_containers

superpoussin22:
last docker version no more use docker -d  sounds better to check dockerd


mboelen:
Thanks, good suggestion.

PR #414, opened 2017-07-04
Firewall check updates

mostafahussein:
This PR will add two checks:

1. Check if CSF is running
2. Check APF (Advanced Policy Firewall) presence along with checking for testing mode status and check if its running


mboelen:
Thanks, they have been merged.

PR #423, opened 2017-07-24
Suhosin is hard linked into PHP on OpenBSD

alobodzinski:


mboelen:
Thanks, it has been merged.

PR #426, opened 2017-07-28
Enhanced detection for LXC and LXC over VM

optyfr:
Modified IsVirtualMachine() to :
- detect correctly LXC container in over-virtualization use case
- correct lscpu hypervisor detection

Added LXC detection code in IsContainer()


mboelen:
Looking good, merged. Thanks!

PR #438, opened 2017-08-13
Do not unset PHPVERSION after detect it

Tom-H-:
For some reason, the variable PHPVERSION is reset at the beginning of tests_php file. So, the tests that use it cannot work (e.g. PHP-2368).


mboelen:
Thanks, good catch :+1:

PR #439, opened 2017-08-16
Added php.ini locations for Ubuntu 16.04LTS

Jsitech:
Lynis does not find php.ini files when executed under Ubuntu 16.04LTS. Added the php.ini file locations for ubuntu 16.04LTS to tests.php.


mboelen:
Thanks, it has been merged

Issue #462, opened 2017-09-14
IsRunning() incorrect match


FlorentCoppint:
Hi,

Lynis is reporting lack of MongoDB authorization (DBS-1820) on a host not running MongoDB server !

After searching, this host is running some commands like : 

`/usr/bin/php -d magic_quotes_gpc=Off -d magic_quotes_runtime=Off -d magic_quotes_sybase=Off /home/www-data/vendor/drush/drush/drush.php --php=/usr/bin/php --php-options= -d magic_quotes_gpc=Off -d magic_quotes_runtime=Off -d magic_quotes_sybase=Off --backend=2 --root=/var/www --yes --uri=default mongodb-advancedqueue-process-corequeue --all --exclude=xxx --timeout=21600`

And you can see in Lynis sources at https://github.com/CISOfy/lynis/blob/master/include/functions#L1250
`FIND=$(${PSBINARY} ${PSOPTIONS} | egrep "( |/)$1" | grep -v "grep")`
This matches my process, even if mongod is not running on this host !

Maybe this function needs to be remastered. I don't have any solution for now.

FlorentCoppint:
Maybe using pgrep command ? https://linux.die.net/man/1/pgrep

mboelen:
Pull request was merged and function was extended with pgrep support. Thanks for reporting this improvement to the function to be more precise in what it matches.

PR #463, opened 2017-09-14
Improved IsRunning() process matching

FlorentCoppint:


FlorentCoppint:
My suggestion to fix #462 
Maybe needs Busybox testing...

mboelen:
Thanks. Will add pgrep support as well for better matching.

PR #466, opened 2017-09-20
Fix displayed result for signing keys check

vins1993:
The result of the signing keys check is saved under KEYS_USED variable,
but SSL_USED was used to present the result to the end user.


mboelen:
Thanks!

PR #470, opened 2017-10-04
remove non printable characters from code

BlessJah:
Non printable ESC character is required to obtain terminal escape
sequence i.e. for changing output color.

Such sequences (especially ESC character) were replaced by command
substitution producing exactly same result (variable value), but using
only "safe" characters.

Use of printf and especialy '\033' or '\0ddd' sequences is described
here:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html#tag_20_94_13

Use of $(command) or command substitution is described here:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03

Verbatim TAB characters were replaced with \t escape sequence as
described to avoid problems with editors silently replacing them or
developer accidentialy messing up the regex.

### DIFF
--- a/include/consts
+++ b/include/consts
@@ -327,33 +327,33 @@ unset LANG
 #
 
     # Normal color names
-    CYAN="[0;36m"
-    BLUE="[0;34m"
-    BROWN="[0;33m"
-    DARKGRAY="[0;30m"
-    GRAY="[0;37m"
-    GREEN="[1;32m"
-    LIGHTBLUE="[0;94m"
-    MAGENTA="[1;35m"
-    PURPLE="[0;35m"
-    RED="[1;31m"
-    YELLOW="[1;33m"
-    WHITE="[1;37m"
+    CYAN="$(printf '\033[0;36m')"
+    BLUE="$(printf '\033[0;34m')"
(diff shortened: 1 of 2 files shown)


mboelen:
On what platforms was this tested?

BlessJah:
The code is compatible with pretty much every POSIX-compliant system (that includes Linux, *BSD, AIX, HP-UX, Solaris and MacOS). I've tested it on linux and will check on FreeBSD and MacOS.

I think we should go step further and conditionally disable colors for non-interactive use. It is possible to check if standard output is a terminal with `test -t 1` or `[ -t 1]`.

Color escape sequences must be supported by terminal emulator (e.g. xterm, Gnome Terminal or PuTTY). I can't recall any meaningful terminal that wouldn't support that, but I always took the feature for granted.

 - https://en.wikipedia.org/wiki/ANSI_escape_code
 - http://pubs.opengroup.org/onlinepubs/009695399/utilities/test.html

mboelen:
Merged and tested. Thanks.

Had a look at -t once, but thought it was not implemented in all shells.

Issue #435, opened 2017-08-08
[FILE-6362] Add check for sticky bit on /var/tmp


CarstenGrohmann:
Hi,

please extend the sticky bit check to /var/tmp additionally to the existing check on /tmp

Regards,
Carsten

mboelen:
Fixed with #473

PR #473, opened 2017-10-08
[FILE-6363] Check for sticky bit on /var/tmp

katkad:
adresses this suggestion https://github.com/CISOfy/lynis/issues/435


mboelen:
Thanks for the pull request, it has been merged

PR #479, opened 2017-10-22
Add additional DirectAdmin paths for php.ini

jamesmacwhite:
My PHP FPM setup is currently not detected by Lynis. I use DirectAdmin and the php.ini for php-fpm via Custombuild 2.0 is installed to the following location:

```/usr/local/php70/lib/php.ini```

The `php70` would vary based on the PHP base version.

https://help.directadmin.com/item.php?id=301

It looks like the default PHP CLI path is already searched, but the CGI and PHP-FPM implementations would not be detected. I've added these paths to the `PHPINILOCS` variable.


mboelen:
Thanks, it has been merged

PR #483, opened 2017-10-29
Avoid /tmp symlink race in unpriviledged scan mode

katkad:
If someone makes:
ln -s /home/$user/.bashrc /tmp/lynis.log

$user being user which uses lynis,
contents of .bashrc are overwritten.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 .bzrignore
 custom.prf
 *.swp
+lynis-report.dat
+lynis.log
--- a/lynis
+++ b/lynis
@@ -236,17 +236,17 @@ Make sure to execute ${PROGRAM_NAME} from untarred directory or check your insta
     # Disable logging if no alternative was provided
     if [ ${PRIVILEGED} -eq 0 ]; then
         if [ -z "${LOGFILE}" ]; then
-            # Try creating a log file in temporary directory
-            if [ ! -f /tmp/lynis.log ]; then
-                touch /tmp/lynis.log
-                if [ $? -eq 0 ]; then LOGFILE="/tmp/lynis.log"; else LOGFILE="/dev/null"; fi
+            # Try creating a log file in current working directory
+            if [ ! -f ./lynis.log ]; then
+                touch ./lynis.log
+                if [ $? -eq 0 ]; then LOGFILE="./lynis.log"; else LOGFILE="/dev/null"; fi
             else
-                LOGFILE="/tmp/lynis.log"
+                LOGFILE="./lynis.log"
             fi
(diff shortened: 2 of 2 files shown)


mboelen:
Thanks for the suggestion. Will have to think about this suggestion. Sure, user 1 could do this and get his/her file overwritten when user 2 runs the tool. But there is not much to gain, except your own file being overwritten with plain data. Using the current working directory is not always possible, as one might have no write access to that directory.

What if we rewrite this by doing a check for the file /tmp/lynis.log. If it is a symlink, then we stop execution of the program. That further reduces any risks involved. What do you think?

mboelen:
We made a change that simply checks for the presence of a symlink on said files. Closing this pull request. Thanks for your contributions!

lkundrak:
Hi. Hope it's all right if I chime in here. I think the fix that was applied is wrong and there's still a security problem that could be exploited without too much effort.

> What if we rewrite this by doing a check for the file /tmp/lynis.log. If it is a symlink, then we stop execution of the program. That further reduces any risks involved. What do you think?

> We made a change that simply checks for the presence of a symlink on said files. Closing this pull request. Thanks for your contributions!

There's still a race. A malicious user can in theory succeed creating a symlink file between the check and the actual write to the file. There are techniques that make the race a lot more likely to hit than one would expect.

> But there is not much to gain, except your own file being overwritten with plain data.

This is an understatement. The shell is pretty tolerant to malformed scripts and it's remarkably easy to craft something that would actually end up executing something malicious.

Overall, using /tmp for anything that is not a temporary file is almost always the wrong. The directory is world-writable and needs *very* careful handling (that basically boils down to mktemp/mkstemp) -- in particular you must not ever use a predictable names there. If your name is predictable, it's a strong indication you should be either accepting the location of the output as a parameter or, if it's a well-known name, one of the /var subdirectories (/var/run, /var/log or /var/lib, depending on characteristics and expected lifetime of the data).

Hope that helps.
Lubo

mboelen:
Hi @lkundrak - Thanks for your feedback and insights.

Storing data in /tmp only happens when the tool is run as a non-privileged account, where it is unknown if one can write to /var/log for example. Both /var/run and /var/lib look not the best location either. I think we will go for the home directory of the user, as the file is considered to be a log and data file. Then it is up to the user to decide what to do with it. New issue has been created: https://github.com/CISOfy/lynis/issues/500

katkad:
Hello, I noticed #500 
what about pid file? https://github.com/CISOfy/lynis/blob/master/lynis#L319

do you know of any case when tilde expansion fails?

PR #484, opened 2017-10-30
README.md: added 'cd lynis' after git clone

icasimpan:
Minor issue but helpful in completing step-by-step usage.


mboelen:
Thanks, that is indeed a small yet important step. Merged!

PR #487, opened 2017-10-30
Check zero size of files correctly

katkad:
from man test:
-s FILE
    FILE exists and has a size greater than zero

-z STRING
    the length of STRING is zero


mboelen:
Thanks, it has been merged

PR #494, opened 2017-11-23
Fix detecting systemd timedate daemon

Shalzz:
timedatectl status output was changed by the commit systemd/systemd@3ec530a

### DIFF
--- a/include/tests_time
+++ b/include/tests_time
@@ -99,7 +99,7 @@
 
         # Check timedate daemon (systemd)
         if [ ! -z "${TIMEDATECTL}" ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for the pull request and your feedback! Seems like a good addition.

As current systemd implementations still use "NTP ....", can you change your commit that it looks for both?
${EGREPBINARY} "^(NTP|System clock) synchronized: yes"`

Then we can merge your contribution :)

Shalzz:
Done!

mboelen:
Merged, thanks!

PR #498, opened 2017-12-09
lynis should not suggest suhosin for PHP 7

### LINKED ISSUE #420: PHP-2379 should not suggest suhosin for php 7
PHP-2379 tries to detect suhosin extension on PHP, and suggest it if not available, the problem is that suhosin for php 7.x (suhosin7) is not available and won't be available for months since it is pre-alpha release (and not actively developed)
I suggest to exclude PHP-2379 when there is only php 7.x installed and running, since It may be more harmful to use suhosin7 than being vanilla

noone-silentrunner:
resolves #420 also adds a warning if suhosin is enabled for PHP 7

### DIFF
--- a/include/tests_php
+++ b/include/tests_php
@@ -65,6 +65,12 @@
                 ${ROOTDIR}opt/alt/php71/etc/php.ini"
 
     PHPINIDIRS="${ROOTDIR}etc/php5/conf.d \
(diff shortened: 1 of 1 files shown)


mboelen:
merged, thanks!

PR #499, opened 2017-12-15
Adding USBGuard to checks for USB Devices.

dvehrs:
The current tests for the usb-storage module and USB Authorization seems a little too close to the all or nothing approach for enabling USB on a system.  While that is appropriate in many situations, there are others where a more fine-grained control of which USB are or are not allowed is needed.  USBGuard provides that in a way that appears to be more user friendly than writing a series of udev rules for USB Authorizations.  

I have created a few tests to determine if USBGuard is installed and if it is configured in an appropriately hardened way.   To do so, I created a new file "tests_usb_devices" to hold these tests and I wonder if it might not be appropriate to move the tests for usb-storage and USB Authorization from tests_storage to it as well.

Also I have not enabled a test name or number so all the suggestions should be disabled until one is assigned.

What do you think?  Is this path worth following or does it potentially create more problems than it solves?

Thanks,
Dave


mboelen:
Hi Dave, thanks for your pull request.

The idea is good, but as the code changes are quite a lot to pull in at the same time, we will have to do more analysis on this pull request. So that is the reason why it is waiting for a bit.

dvehrs:
Thanks for the heads up.  I completely understand and let me know if there is anything I can do to help.

mboelen:
Thanks. Initial merge performed, some rework is needed (e.g. numbering, rename to 'usb' group). Let's do that in steps together. Then when everything works as expected, we turn on the 'usb' group in one of the upcoming releases.

dvehrs:
Sounds good but a little intimidating.  I'm not sure I understand enough of the big picture items that need to be done so I may need a little guidance as for which to work on next.    How much of idea was good?  Was it just adding tests for usbguard or does it include moving the existing tests for usb-storage and usb authorization into tests_usb?  And what does that mean for numbers for those tests?  

if we do move all three to tests_usb would it be appropriate to number them?
  usb-storage == USB-0100
  usb authorization == USB-0200
  usbguard == USB-0300

This leaves some room in between for tests we develop in the future and there may be ways to chain the tests so that any Suggestions make sense and don't overlap.  For example if usbguard is installed then it may not make sense to suggest that usb-storage be disabled, but on the other hand if neither are installed then maybe a suggestion that lists the options available would be best.  

Or am I over-thinking this?

Issue #501, opened 2018-01-04
BitDefender check out of date


opoplawski:
The BitDefender check looks for the "bdagentd" process.  But on my BD systems I have:

MacOS:
```
   80   ??  SNs    6:10.55 /Library/Bitdefender/AVP/enterprise.bundle/epagd
  179   ??  Ss    32:08.02 /Library/Bitdefender/AVP/antivirus.bundle/BDUpdDaemon
  180   ??  Ss    94:02.56 /Library/Bitdefender/AVP/antivirus.bundle/BDLDaemon
  181   ??  Ss     3:21.34 /Library/Bitdefender/AVP/common.bundle/BDCoreIssues
  571   ??  UN    30:05.33 /Library/Bitdefender/AVP/EndpointSecurityforMac.app/Contents/MacOS/EndpointSecurityforMac
```

Linux:
```
root     13069     1  0 Jan02 ?        00:00:21 /opt/BitDefender/bin/bdregd -s
root     13075     1  0 Jan02 ?        00:00:04 /opt/BitDefender/bin/bdlogd -s
root     13082     1  0 Jan02 ?        00:00:59 /opt/BitDefender/bin/epagd -s
root     13089     1  0 Jan02 ?        00:07:29 /opt/BitDefender/bin/bdepsecd -s
root     13102     1  0 Jan02 ?        00:04:38 /opt/BitDefender/bin/bdsrvscand -s
root     13209     1  0 Jan02 ?        00:00:29 /opt/BitDefender/bin/bdmond -s
root     13217     1  0 Jan02 ?        00:03:39 /opt/BitDefender/bin/bdlived -s
```

perhaps looking for "epagd" might be better?

mboelen:
Sounds like a good match for both. Like to create a pull request for it?

mboelen:
@opoplawski do you want to create a pull request with your suggestion?

opoplawski:
Not opposed to doing it.  Not sure I'll find time....

mboelen:
Shouldn't take long and we would really appreciate it, as it shows more contributors to the project.

If it takes too long or someone else swoops in, we let you know :)

mboelen:
Thanks to @Jsitech a pull request was made. Closing issue.

PR #503, opened 2018-01-12
Changed Process name search for Bitdefender test.

Jsitech:
Changed the Process name Search for bitdefender since it do matches MacOS and Linux. bdagentd to epagd.  Issue "BitDefender check out of date #501"


mboelen:
Thanks!

PR #505, opened 2018-01-14
VBoxService running is another VirtualBox clue

mslifcak:
Add a check for process VBoxService , observed to be running in a vagrant VM that uses VirtualBox 5.1.22.


mboelen:
Merged with several other minor code enhancements. Thanks!

PR #507, opened 2018-01-14
Pin isrunning

mslifcak:
Problems with existing version:
1. Searches for exim running;  Does not see /sbin/exim4, the instance of exim (Ubuntu).
2. PSOPTIONS is set at most one time;  Incorrect for subsequent IsRunning calls.
3. The --full option is not needed, since the search parameter is quoted in each use.

IsRunning uses are re-factored by removing the "--full" argument.
The side effect of PSOPTIONS are eliminated when the variable is removed.

With this patch, IsRunning searches a temporary file. The first match for the search is displayed.
The temporary file is populated with the output of several "ps" command variants.
Errors are ignored for the "ps" command variants that are not supported on the system under test.  This is necessary since there is no one "ps" usage that finds all the processes for every OS.
 "ps" commands known to produce better results for some OS can be inserted in the subshell used when the temporary file is created.
Simple. Not elegant. Effective.


mslifcak:
See pin-find-exim4 patch (one line) that more succinctly meets the need to find the exim process when it has a slightly different name (exim4).

PR #508, opened 2018-01-16
Create GR

alphaomicron:
Simply added Greek language


mboelen:
Thanks, it has been merged.

PR #509, opened 2018-01-16
restore UsePrivilegeSeparation to list of ssh daemon checks

mslifcak:
This configuration item is still in play on modern SSH (Ubuntu,RHEL 7).  it may be on by default, but in my opinion, it should still be tested.

It was active in 2.5.1, but by the time 2.5.7 came around, it was missing.  This patch restores the value.

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -127,6 +127,7 @@
                 StrictModes:YES,,NO:=\
                 TCPKeepAlive:NO,,YES:=\
                 UseDNS:NO,,YES:=\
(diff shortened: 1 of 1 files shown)


mboelen:
Note: this was removed with commit https://github.com/CISOfy/lynis/commit/b3dffbf750ff26f92e78fb5f32b068e5c6c6c6b8

mboelen:
Change merged, as for now many systems will still use this setting. When OpenSSH 7.5 is the standard for most systems, then this setting can be removed.

PR #510, opened 2018-01-17
sort BIN_PATHS before process

mslifcak:
sorting BIN_PATHS before traversing each one will produce results that are repeatable.  This isn't a big deal but it is nice to have when comparing logs to establish trends.


mboelen:
Thanks, merged

PR #511, opened 2018-01-18
touchup BOOT_LOADER_SEARCHED setting for BOOT-5106 and BOOT-5108

mslifcak:
The SEARCHED flag seems it should be set for these tests.


mboelen:
Thanks, merged!

PR #512, opened 2018-01-18
also find exim4 in MAIL-8802

mslifcak:
Checks for exim only if exim4 is not running.

exim4 is known to be Ubuntu 16's exim...


mboelen:
Merged, thanks!

PR #514, opened 2018-01-26
tests_usb updates

dvehrs:
OK, I've got a couple updates to tests_usb to see if I'm on the right path.

So far:
1.  Moved usb-storage and usb authorization tests from tests_storage to tests_usb
2.  Added simple tests to quiet recommendations for usb-storage if usbguard is installed.
3. Changed test from usb_devices to usb in lynis

I've only done some very preliminary testing with this on one system so it definitely needs more testing but I wanted to find out if I was headed in the right direction before I went too far.

What do you think?  Is this a workable solution?

Should we move the firewire tests from tests_storage into a tests_firewire as we are doing with the usb tests?


mboelen:
Merged, will make some minor changes to the code and numbering.

This looks fine this way. Keep pulling in small changes, then we review, you merge, make changes, etc. 
Let's leave firewire as-is now, to keep focus on just USB. Otherwise, the renumbering will cause some issues.

Suggestion: Before creating a pull request, first merge your code with the latest master. Now changes like sysctl are part of your pull request. This makes it harder to review and filter out the actual changes.

PR #519, opened 2018-02-08
Pin db sync

mslifcak:
There are several tests that have minor differences between the testname in the comment, the Register statement, and db/tests.db. 
There is one duplicate db record.
There are nine tests that have no db record.  New db records are created using the Register arguments for testname, os, and description.  The Group is formed from the filename that holds the test.  The CORE-1000 is given to the system_integrity group as it seems the best fit.

The db records are sorted.  No FINT,PROC lines were changed, except for their position.

Given that the USB tests are migrating from STRG, these changes should be scrutinized carefully.


mboelen:
Thanks, merged!

Issue #515, opened 2018-01-26
FIRE-4508: null byte error


mboelen:
OS: Fedora 26

[DEBUG] Performing test ID FIRE-4508 (Check used policies of iptables chains)
    - Checking iptables policies of chains                    [ FOUND ]
/usr/share/lynis/include/tests_firewalls: line 118: warning: command substitution: ignored null byte in input

mboelen:
Resolved with recent commit

PR #520, opened 2018-02-13
Solved Null byte bash Warning Issue #515

Jsitech:
Solved Null byte bash warning referenced on Issue #515 
FIRE-4508: null byte error 

Similar issue was present on line 74 test_boot_services. Was resolved here as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Lynis Changelog
 ===============
 
+Lynis 2.6.3 (not released yet)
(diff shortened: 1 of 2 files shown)


mboelen:
Can you run a 'git pull' in your tree? Now there is a whole list of commits attached to this pull request :)

Jsitech:
This pull request got messed up and could not find a good workaround. Synchronized both Repos and submitted a new Pull Request.

PR #531, opened 2018-03-12
delayed is a legacy synonym

konstruktoid:
Suggesting `Compression (YES --> (DELAYED|NO))` is not necessary since delayed is `a legacy synonym for yes` according to SSHD_CONFIG(5).


mboelen:
Thanks, it has been merged

PR #535, opened 2018-03-17
DNS-1600 Check for DNSSEC validation

kisst:
- new test category DNS
 - first test in the category DNS-1600
 - update changelog


mboelen:
merged, will be disabled initially as our own domain needs to be configured (or a test domain)

PR #539, opened 2018-04-12
Initial Tests for Exim

dvehrs:
Here are a few tests for Exim.  Needs to be expanded but I was wondering if this is a good start.  

First test is to determine what type of Exim configuration the user has selected.  This is one of five options:  local only, internet host, smarthost, satellite, or unconfigured.  These five options are based on the choices given to a user during installation on Debian but they do not rely on Debian's update-exim4.conf file rather they are determined from configuration options that are set in Exim's configuration.  All tests rely on "exim -bP" to extract the current configuration. 

Second test is to verify if TLS is enabled, basically if Exim is configured to advertise TLS to anyone then it is considered enabled.

Third test is to verify that a TLS Certificate was configured.

Fourth test is to verify that a TLS Private Key was configured.

The last two tests are slightly different for two options, they are configured one way for internet host or smarthost and a different way for satellites.  

This is only a beginning for Exim tests and there are more to add but I was curious if I'm headed in a reasonable direction before I invest too much time into it.  What do you think?  Do these tests give us a good framework to build off?

Dave


mboelen:
Code looks good, new additions can be submitted when you have the chance.

PR #553, opened 2018-05-29
[TIME-3160] Extend check for step-tickers file for RedHat

CarstenGrohmann:
Hi Michael,

on RedHat if the step-tickers file exists even it it's empty, the ntpdate start script uses the servers listed in ntp.conf for the initial time synchronization.

Regards,
Carsten


mboelen:
Thanks, it has been merged

PR #557, opened 2018-06-22
Updated tests_ssh, removed extra ssh in the test 

aram535:
Seems like in the patch there was an extra 'ssh' added in the command line, which is breaking the ssh tests.  Removing the ssh keyword... -T -C ... fixes the problem.


mboelen:
Thanks, it has been merged

PR #560, opened 2018-07-12
Fix for umask check in case of multiple files

gkrystev:
In case when umask is checked in multiple files and in some of the files except the last one a weak umask is found, the tool reports weak mask for the rest of the files. In the example bellow, the weak umask is only in /etc/csh.cshrc. However, the check /etc/profile is reported weak as well.

Expected:

Checking default umask values
Checking default umask in /etc/bashrc [ OK ]
Checking default umask in /etc/csh.cshrc [ WEAK ]
Checking default umask in /etc/profile [ OK ]
Actual:

Checking default umask values
Checking default umask in /etc/bashrc [ OK ]
Chhhhecking default umask in /etc/csh.cshrc [ WEAK ]
Chhhhecking default umask in /etc/profile [ WEAK ]

### DIFF
--- a/include/tests_shells
+++ b/include/tests_shells
@@ -235,9 +235,9 @@
     Register --test-no SHLL-6230 --weight H --network NO --category security --description "Perform umask check for shell configurations"
     if [ ${SKIPTEST} -eq 0 ]; then
         FOUND=0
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, it has been merged!

Issue #561, opened 2018-07-15
apache24 detection does not work properly on FreeBSD


rahra:
Apache24 modules on FreeBSD and config files reside in ....../apache24 subdirectories. The paths should be added to `test_webservers`:
(tested on stable FreeBSD 11.2)
```
diff --git a/include/tests_webservers b/include/tests_webservers
index d54d279..b9ca72c 100644
--- a/include/tests_webservers
+++ b/include/tests_webservers
@@ -29,7 +29,7 @@
     # Reset Apache status
     APACHE_INSTALLED=0
     APACHE_MODULES_ENABLED_LOCS="${ROOTDIR}etc/apache2/mods-enabled"
-    APACHE_MODULES_LOCS="${ROOTDIR}etc/httpd/modules ${ROOTDIR}opt/local/apache2/modules ${ROOTDIR}usr/lib/apache ${ROOTDIR}usr/lib/apache2 ${ROOTDIR}usr/lib/httpd/modules ${ROOTDIR}usr/libexec/apache2 ${ROOTDIR}usr/lib64/apache2 ${ROOTDIR}usr/lib64/apache2/modules ${ROOTDIR}usr/lib64/httpd/modules ${ROOTDIR}usr/local/libexec/apache ${ROOTDIR}usr/local/libexec/apache22"
+    APACHE_MODULES_LOCS="${ROOTDIR}etc/httpd/modules ${ROOTDIR}opt/local/apache2/modules ${ROOTDIR}usr/lib/apache ${ROOTDIR}usr/lib/apache2 ${ROOTDIR}usr/lib/httpd/modules ${ROOTDIR}usr/libexec/apache2 ${ROOTDIR}usr/lib64/apache2 ${ROOTDIR}usr/lib64/apache2/modules ${ROOTDIR}usr/lib64/httpd/modules ${ROOTDIR}usr/local/libexec/apache ${ROOTDIR}usr/local/libexec/apache22 ${ROOTDIR}usr/local/libexec/apache24"
     NGINX_RUNNING=0
     NGINX_CONF_LOCS="${ROOTDIR}etc/nginx ${ROOTDIR}usr/local/etc/nginx ${ROOTDIR}usr/local/nginx/conf"
     NGINX_CONF_LOCATION=""
@@ -42,7 +42,7 @@
         ${ROOTDIR}opt/apache \
         ${ROOTDIR}usr/local/apache ${ROOTDIR}usr/local/apache2 \
[...]

PR #562, opened 2018-07-20
added path information Apache 2.4 on FreeBSD

rahra:
As mentioned in issue #561, the default installation of Apache 2.4 from the ports tree on FreeBSD goes to /..../apache24 folders.
This pull-request contains the patch for this.
Tested on FreeBSD 11.2 and 10.4


mboelen:
Thanks, it has been merged!

PR #568, opened 2018-08-01
Update helper_audit_dockerfile

superpoussin22:
To support LABEL maintainer="toto" and LABEL maintainer "toto"
correct syntax from docker is LABEL maintainer="xxxxxxxxxxxxxxx"


mboelen:
Thanks, it has been merged!

PR #574, opened 2018-08-15
A few more Exim4 settings

dvehrs:
OK, here are a few more Exim4 configuration settings that I've been playing with that seem to be relevant for Lynis.  There may be some confusion with the "Try Verify Hosts" and "Verify Hosts" but so far I haven't been able to come up with a simple way to differentiate them in the report and making it clear.   It's almost as if the user needs to be aware of the difference they have for Exim's configuration.  

So what do you think?  Does this work well enough that it gives us something to build on going forward?

Dave


mboelen:
Dave,

I can't commit your changes as your pull request includes many old commits and pull request merges. So I don't even know where to look for the specific changes.

Please create a new pull request by first synchronizing your copy with the master. Then apply the changes and create a pull request. This way it should be 'clean' and only have relevant changes in the pull request for me/us to review.

Thanks,

Michael
(closing this one)

PR #575, opened 2018-08-25
detect if latest TAG is used in dockerfile

superpoussin22:
that's always better to specify the version, latest desn't mean latest version 
it can be a suggestion if you prefer


mboelen:
Thanks, it has been merged.

PR #578, opened 2018-09-01
fix opensuse os detection with os-release

wagner-certat:
openSUSE uses the os-release file too

I'm not sure what you want to have in OS_NAME vs LINUX_VERSION


superpoussin22:
perhaps a generic case for openSUSE would be better ?

mboelen:
Thanks, it has been merged. Let's have a look in the future to merge things, so it can be more generic.

PR #586, opened 2018-10-02
Update tests_ports_packages

compermisos:
change " " <- space for [[:blank:]] (to clean all pausible spaces)


compermisos:
only for info, some Yum versions print "short space" inside the string in this test (additional to normal space).

PR #589, opened 2018-10-10
Added support for TOMOYO Linux Mandatory Access Control

deonspengler:
Hi Lynis community,

If there is any changes that needs to be made please let me know.


deonspengler:
Thanks. How would we ago about changing the CUST-0001 and CUST-0002 to something along the lines of MACF-XXXX

mboelen:
They have new identifiers now.

PR #593, opened 2018-10-23
osdetection: ignore quotes in OS_ID

wagner-certat:
Strip the double quotes from OS_ID too, currently it's only done for all the other fields.


mboelen:
Thanks, it is merged!

PR #595, opened 2018-10-28
[SHLL-6230] Add etc/bash.bashrc.local to umask check

### LINKED ISSUE #563: umask check for bash also needs to look at bash.bashrc.local
My OS welcomes me like this:

```
# /etc/bash.bashrc for SUSE Linux
#
# PLEASE DO NOT CHANGE /etc/bash.bashrc There are chances that your changes
# will be lost during system upgrades.  Instead use /etc/bash.bashrc.local
# for bash or /etc/ksh.kshrc.local for ksh or /etc/zsh.zshrc.local for the
# zsh or /etc/ash.ashrc.local for the plain ash bourne shell  for your local
# settings, favourite global aliases, VISUAL and EDITOR variables, etc ...
```

Accordingly I'm setting the umask for bash in ```/etc/bash.bashrc.local```.

This isn't seen though:

```
[+] Shells
  - Checking shells from /etc/shells
    Result: found 23 shells (valid shells: 9).
    - Session timeout settings/tools                          [ NONE ]
  - Checking default umask values
    - Checking default umask in /etc/bash.bashrc              [ NONE ]
    - Checking default umask in /etc/csh.cshrc                [ NONE ]
    - Checking default umask in /etc/profile                  [ NONE ]
```

(I think this one only takes effect from non-login shells, while the rest is covered via a working ```/etc/profile.d/umask.sh```)

katkad:
fixes #563

### DIFF
--- a/include/tests_shells
+++ b/include/tests_shells
@@ -231,7 +231,7 @@
 #
     # Test        : SHLL-6230
     # Description : Check for umask values in shell configurations
-    SHELL_CONFIG_FILES="${ROOTDIR}etc/bashrc ${ROOTDIR}etc/bash.bashrc ${ROOTDIR}etc/csh.cshrc ${ROOTDIR}etc/profile"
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, it has been merged!

PR #601, opened 2018-11-12
Fix error printed to screen when run from shell without "command"

theycallhimpat:
Issue is present on busybox ash shell without "command" builtin. Fix
issue by redirecting errors to /dev/null


mboelen:
Thanks, looks good. It has been merged!

mboelen:
Thanks, looks good. It has been merged!

PR #602, opened 2018-11-12
Fix printed error when wget comes from busybox

theycallhimpat:
Busybox's wget does't provide the -V parameter to get the version, so
redirect stderr to /dev/null to hide the printed error message


mboelen:
Thanks, it has been merged

PR #603, opened 2018-11-22
Fix MacOS Mojave detect pattern

kehao95:


mboelen:
Thanks, it has been merged!

PR #633, opened 2019-02-01
[BOOT 5177] Simplify service filter & support multiple periods in names

chr0mag:
The first commit fixes a corner case where _systemd_ service names contain multiple periods and were getting truncated after the first one. Using '-F.service' as the _awk_ filter instead of '-F.' ensures the entire service name up to '.service' is captured. Grepping the source tree didn't turn up any further instances of this.

The 2nd commit simplifies the service filtering to use _systemctl_ switches instead of an additional pipe to _awk_.


mboelen:
Great improvements, thanks. It has been merged and will be available in the next release.

PR #634, opened 2019-02-02
[BOOT-5117] Add systemd-boot bootloader detection

chr0mag:
Adds a test to detect systemd-boot. The 'bootctl' binary is also
added as this is the utility used to inspect the systemd-boot
configuration.

This test is only executed if systemd is installed, the bootctl
utility exists and the system is booted in UEFI mode.


mboelen:
Thanks @chr0mag! Change looks good and has been merged.

PR #637, opened 2019-02-05
[AUTH-9252] Sudo configuration file/folder check improvements

### LINKED ISSUE #600: AUTH-9252 should check files in sudoers.d
sudo uses config snippets in `sudoers.d` in addition to its main `sudoers` config file. These snippets may be just as sensitive as the main config file is. AUTH-9252 currently checks `sudoers` only.

This has been observed with Lynis 2.7.0 (but most likely has always been this way).

Lynis should be checking files in `sudoers.d` in addition to the `sudoers` file.

chr0mag:
First commit adds permissions checks for files in the configuration drop-in directory (eg. /etc/sudoers.d) if it exists and closes #600 .

Second commit adds permissions checks for the drop-in directory if it exists. Previously only files were checked.

Third commit adds file & directory ownership checks to ensure all sudo configuration files/folders are owned by UID=0 and GID=0.

Also includes minor changes to phrasing and changing all references of 'ls' to 'LSBINARY'.

This has been tested on Linux (Archlinux). Tests included checking (and finding) bad drop-in directory permissions and bad permissions/ownership of files within the drop-in directory.

Needs testing on non-Linux hosts, specifically *BSDs which I understand can have multiple accounts with a UID of 0.

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -483,34 +483,36 @@
 #################################################################################
 #
     # Test        : AUTH-9252
-    # Description : Check permissions for sudo configuration files
+    # Description : Check ownership and permissions for sudo configuration files
     if [ ! -z "${SUDOERS_FILE}" ]; then PREQS_MET="YES"; else PREQS_MET="NO"; fi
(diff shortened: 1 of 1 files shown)


mboelen:
Looks good, will be merged. Thanks!

PR #641, opened 2019-02-15
better OpenBSD support

jirib:
- doas is OpenBSD sudo alternative
- OpenBSD uses init and /etc/rc, /etc/rc.d, thus let's call its service manager bsdrc not to see UNKNOWN
- OpenBSD uses rcctl for start/enable daemons (well it lists also non-daemons thus filtered out via grep)
- check OpenBSD rc scripts and configs for not being world readable
- OpenBSD uses clang and gcc as compiler, both present via cc and clang (amd64) and gcc for GNU cc
- php ini and configuration on OpenBSD tunning


mboelen:
Hi @jirib - Thanks for the work on this and the changes. Well done, it has been merged.

PR #650, opened 2019-03-03
[KRNL-5788] Fix false positive warning on missing /vmlinuz

### LINKED ISSUE #649: [KRNL-5788] False positive warning on missing /vmlinuz
The current code in test KRNL-5788 assumes all Debian systems have a symlink /vmlinuz. This assumption is wrong, however, expecially for architectures that use different bootloaders such as armhf and U-Boot. Such systems (set up with the official Debian installer) may not have /vmlinuz but only /boot/vmlinuz.

I'm attaching a patch that fixes this by falling back to checking /boot/vmlinuz if /vmlinuz is not found.
[KRNL-5788-Fix-false-positive-warning-on-missing-vmlinuz.txt](https://github.com/CISOfy/lynis/files/2922156/KRNL-5788-Fix-false-positive-warning-on-missing-vmlinuz.txt)

silentcreek:
Not all architectures use a /vmlinuz symlink in Debian. For instance,
armhf systems may only provide a symlink in /boot/vmlinuz. Fall back to
testing /boot/vmlinuz if /vmlinuz is not found.

Fixes #649

Signed-off-by: Timo Sigurdsson <public_timo.s@silentcreek.de>

### DIFF
--- a/include/tests_kernel
+++ b/include/tests_kernel
@@ -374,10 +374,17 @@
         LogText "Test: Searching apt-cache, to determine if a newer kernel is available"
         if [ -x ${ROOTDIR}usr/bin/apt-cache ]; then
             LogText "Result: found ${ROOTDIR}usr/bin/apt-cache"
-            LogText "Test: checking readlink location of ${ROOTDIR}vmlinuz"
-            if [ -f ${ROOTDIR}vmlinuz ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks @silentcreek - It has been merged.

PR #653, opened 2019-03-08
Quote WORKDIR directory existence in lynis

ee7git:
Quote both "${WORKDIR}/include" and "${WORKDIR}/db" directory existence check in if-statements to avoid "Fatal error: can't find directory" in case of a space-character in the ${WORKDIR} path.


mboelen:
Thanks for the pull request and your first contribution! It has been merged.

Issue #608, opened 2018-12-02
AUTH-9282 : Search password protected accounts without expire date


Capashenn:
**Describe the bug**
The test AUTH-9282 is always OK on non RedHat based distribution, even if accounts doesn't have expire date.

**Version**
 - Distribution : non RedHat based
 - Lynis version : All

**Expected behavior**
This test should report a suggestion if there are protected accounts with no expire date.

**Output**
```
passwd -a -S
root P 09/07/2014 0 99999 7 -1
```

lynis.log :
```
[16:13:58] Performing test ID AUTH-9282 (Checking password protected account without expire date)
[16:13:58] Test: Checking Linux version and password expire date status
[16:13:58] Result: all accounts seem to have an expire date
```

**Additional context*
This problem is caused by the [Register](https://github.com/CISOfy/lynis/blob/0308823966cbcde0f3ee47597f9a0b1b8149ff4c/include/tests_authentication#L736) function which resets the FIND variable.
This issue could be elsewhrere in lynis, i didn't check.

Issue #621, opened 2018-12-12
[SSH-7402] SSH_DEAMON_OPTIONS_FILE empty


Capashenn:
**Describe the bug**
I know I'm a little late but in OpenSSH releases before 5.1 the "-T" option didn't exist, so the file SSH_DEAMON_OPTIONS_FILE is always empty and SSH-7408 test never works

**Version**
 - Distribution RHEL 5
 - Lynis version 2.7.1

**Expected behavior**
Lynis should parse /etc/ssh/sshd_config if -T option doesn't exist in OpenSSH

**Output**
```
[root@localhost ~]# sshd -T
sshd: illegal option -- T
OpenSSH_4.3p2, OpenSSL 0.9.8e-fips-rhel5 01 Jul 2008
```

**Additional context**
related to #159

mboelen:
Thanks for reporting.

As this is a really old version, it is not worth the effort of changing the test at this point. Even RHEL 5 itself is no longer supported by the vendor and ended in [2015](https://access.redhat.com/support/policy/updates/errata).

Capashenn:
Although there are still many redhat 5 in some industries, I agree with you.
So, it would make more sense to add a prerequisite on the OpenSSH version for this test, if the version is lower than 5.1, we don't do this test.

Issue #620, opened 2018-12-08
[FINT] AIDE - Ensure database exists


Capashenn:
**Is your feature request related to a problem? Please describe.**
Lynis should ensure that the database exists

**Describe the solution you'd like**
lynis should check that the database file exists because if it doesn't, aide always fail

```
# aide --check
Couldn't open file /var/lib/aide/aide.db.gz for reading
```

A piece of code to help (works with CentOS/Debian): 
```
DBDIR=$(grep ^database= ${AIDECONFIG} | sed "s/.*://")
if case ${DBDIR} in @@*) ;; *) false;; esac; then
    DBDIR=$(echo ${DBDIR} | sed "s#.*}#$(grep  "@@define.*DBDIR" {AIDECONFIG} | awk '{print $3}')#")
fi
if [ -s ${DBFILE} ]; then
    LogText "Result: found aide database ${DBDIR}"
fi
```

**Required changes**
Add a new tests FINT-XXXX to lynis

**Additional context**
This test could also be added for other integrity softwares but i don't know them as well as AIDE

mboelen:
Sounds like a good suggestion and can be easily checked. Can you turn this into a pull request?

Suggested test ID: FINT-4316

Capashenn:
Of course, I can do that.
As soon as I have a moment to spare I will code this new test.

mboelen:
Implemented by @Capashenn with commit https://github.com/CISOfy/lynis/pull/623

mboelen:
This was already implemented, but not closed yet. So closing issue now.

Issue #611, opened 2018-12-06
[CORE-1000] Binaries search and symbolic links


Capashenn:
**Describe the bug**
During the CORE-1000 test, if Lynis finds symbolic links then Lynis does not use the binaries during the audit, which means that some tests fail while the binaries are present on the machine.

**Version**
 - Distribution : mandrake 2010
 - Lynis version 2.1.7

**Expected behavior**
Lynis should use binaries even if they are symbolic links

**Output**
```
FIND=$(ls -p ${SCANDIR} | grep -v '/$')
awk@
lsmod@
```

mboelen:
As you are using a very old version of Lynis, can you upgrade to the last one 2.7.0 and see if detection is improved?

Capashenn:
Excuse me, I was wrong when I wrote the lynis version.
I use version 2.7.1 for my tests.

mboelen:
I had a fresh look at this issue to see why ls would display a '@' at the first place. It looks like non-standard behavior. Do you still have this Mandrake system? Can you have a look if it uses an alias for ls?

mboelen:
Closing issue, as I can't reproduce it and as part of the work on the major release 3.x. Seems that on all modern Linux distributions the existing routine does work correctly, also when it encounters symlinks.

PR #657, opened 2019-03-21
Fix issues #608 #611 #620 #621

Capashenn:
Fix issues #608 #611 #620 #621


mboelen:
Thanks for your work, it is appreciated!

Please split these changes into individual pull requests. Otherwise, the review will be much harder and some changes might not be merged when there are blocking issues.

Have to close this pull request for the reasons above.

PR #669, opened 2019-03-25
fix issues #608 and #611

Capashenn:
fix issues #608 and #611

### DIFF
--- a/include/binaries
+++ b/include/binaries
@@ -92,8 +92,8 @@
                     for FILENAME in ${FIND}; do
                         COUNT=$((COUNT + 1))
                         if case ${FILENAME} in *@) ;; *) false ;; esac; then
-							FILENAME=${FILENAME%?}
-						fi
+                            FILENAME=${FILENAME%?}
+                        fi
                         BINARY="${SCANDIR}/${FILENAME}"
                         DISCOVERED_BINARIES="${DISCOVERED_BINARIES}${BINARY} "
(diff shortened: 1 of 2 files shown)


mboelen:
This patch uses bash instead of POSIX-safe scripting. For that reason, we can't merge it and have to close it.

Capashenn:
Same here, could you explain to me where the bashism is ?

Tested with `dash `and `checkbashisms`.

mboelen:
My apologies, you are right, this is valid code. The `@@*` and `false` looked unusual to me.

Unfortunately, the code does not work on systems like FreeBSD as there is no '@' used for symlinks. Your patch also includes an unneeded rename of the FIND variable into FIND1. While I do understand the thought, the reuse of variables is preferred. So that is why FIND is typically used and sometimes FIND2 is added to the mix.

Capashenn:
I don't really understand your explanation, I just tested the code on FreeBSD and it worked perfectly well.
Indeed, if there is no '@' used for symlinks on FreeBSD, it never goes in the if statement, so there is no problem.
I made this because on some old systems `ls -p` act exactly the same as `ls -F` which cause some errors.

Moreover, as i already explain in issue #608, the [Register function](https://github.com/CISOfy/lynis/blob/256bc1da0fec522dc6cce84bba499416c367f6d8/include/tests_authentication#L766) break the FIND variable and the test AUTH-9282 **never** works.

PR #670, opened 2019-03-25
fix issue #620

Capashenn:
fix issue #620

### DIFF
--- a/include/tests_file_integrity
+++ b/include/tests_file_integrity
@@ -99,6 +99,36 @@
 #
 #################################################################################
 #
+    # Test        : FINT-4316
+    # Description : Check if AIDE database exist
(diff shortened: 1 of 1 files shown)


mboelen:
This patch uses bash shell scripting instead of POSIX-safe script. For that reason, it can't be merged.

`if case ${AIDE_DB} in @@*) ;; *) false;; esac;`

Capashenn:
Could you explain to me where the bashism is ?

I carrefully read the [POSIX specification](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html) and didn't find anything relevent

I tested the script with dash and it worked perfectly well.
Moreover, i used `checkbashisms `and no error reported.

mboelen:
My apologies, you are right, this is valid code. The `@@*` and `false` looked unusual to me.

mboelen:
Closing, see remark on related issue https://github.com/CISOfy/lynis/issues/611

PR #671, opened 2019-03-25
fix issue #621

Capashenn:
fix issue #621

This change removes the patch number at the end of the version and deletes a carriage return if it is part of the output.


mboelen:
Thanks, it has been merged! I made a small additional change that combines both removing a ',' and any control characters instead of just the carriage return (\r).

PR #672, opened 2019-03-25
fix FILE-6374

Capashenn:
fix FILE-6374


mboelen:
Had to retract the changes, as it actually does not show the details correct on a test Ubuntu system. On Fedora I saw no difference in output.

So it looks like we have differences in configuration files. Do you have a particular target OS that failed with the original test in place?

Capashenn:
This problem should happen with any OS.

Actually, the problem is with line like : 
`dev/mapper/centos-root /                       xfs     defaults        0 0`

if you replace `\n` with a space in `FOUND_FLAGS`, the if statement `if [ "${FOUND_FLAGS}" = "defaults" ]; then` can't work because `"${FOUND_FLAGS}" = "defaults "` (notice the space after defaults)

Capashenn:
Could you please show me a fstab entry that doesn't work ?

I'd like to see if i can fix this.

PR #673, opened 2019-03-25
fix SHLL-6220 description

Capashenn:
Corrected SHLL-6220 test description


mboelen:
Good catch, it has been merged

Issue #666, opened 2019-03-21
[HOME] Ensure users' home directories permissions are 750 or more restrictive


Capashenn:
**Is your feature request related to a problem? Please describe.**
Group or world-writable user home directories may enable malicious users to steal or
modify other users' data or to gain another user's system privileges.

**Describe the solution you'd like**
Check users home directories permissions

**Required changes**
Add new test HOME-XXXX

mboelen:
Closing with your pull request merged

Issue #667, opened 2019-03-21
[HOME] Ensure users own their home directories


Capashenn:
**Is your feature request related to a problem? Please describe.**
Since the user is accountable for files stored in the user home directory, the user must be
the owner of the directory

**Describe the solution you'd like**
Check users home directories ownership

**Required changes**
Add new test HOME-XXXX

mboelen:
Closing with your pull request merged

PR #675, opened 2019-03-25
fix issues #666 #667

Capashenn:
fix issues #666 #667


mboelen:
Thanks, it is merged!

Issue #659, opened 2019-03-21
[KRNL-5820] incomplete test


Capashenn:
**Describe the bug**
This test should also check presence of /etc/security/limits.d/* files.

Moreover, this test only ckeck for : 
`* soft core 1`
`* hard core 1`
while the soft/hard core value could be anything else.

**Version**
 - Distribution : All
 - Lynis version : 2.7.1

**Expected behavior**
check in /etc/security/limits.conf and in /etc/security/limits.d/* for : 
`* soft core 0`
`* hard core 0`
Any other value means core dumps are enabled (even if those lines don't exists because ulimit settings can be set on the system)

mboelen:
Closing, with your pull request merged

PR #676, opened 2019-03-25
fix issue #659

Capashenn:
fix issue #659


mboelen:
Thanks, it has been merged!

PR #687, opened 2019-03-26
Check echo via xinetd INSE-8140 INSE-8142

Capashenn:
Check echo via xinetd INSE-8140 INSE-8142
Depend on #686

PR #688, opened 2019-03-26
Check time via xinetd INSE-8144 INSE-8146

Capashenn:
Check time via xinetd INSE-8144 INSE-8146
Depend on #687


mboelen:
Same changes were already merged, closing this pull request

PR #689, opened 2019-03-26
Check tftp via xinetd INSE-8148

Capashenn:
Check tftp via xinetd INSE-8148
Depend on #688 
Fix issue #663

PR #690, opened 2019-03-26
 Check rsync via xinetd INSE-8150

Capashenn:
Check rsync via xinetd INSE-8150
Depend on #689


mboelen:
Thanks - rsync added to INSE-8116 using commit https://github.com/CISOfy/lynis/commit/3f042353535e12972449c06a8193ebfd467349f8

PR #697, opened 2019-03-26
Add tests INSE-8340 INSE-8342 (talk)

Capashenn:
Add tests INSE-8340 INSE-8342 (talk)
Depend on #696

PR #698, opened 2019-03-26
Add tests INSE-8350 INSE-8352 (LDAP)

Capashenn:
Add tests INSE-8350 INSE-8352 (LDAP)
Depend on #697


mboelen:
Thanks for submitting this particular change. While it is generally a good idea to remove unused components, LDAP is generally not something that is activated and forgotten. As part of the preparation for the major 3.x release, this change will not be applied at this moment.

PR #713, opened 2019-04-29
Update tests_authentication - AUTH-9402

bcs016:
Update AUTH-9402, change the uid name to check in etc/passwd file when device is a QNAP


mboelen:
Thanks, it has been merged!

PR #718, opened 2019-05-14
Add danish language support

Patark:
Add danish language file


mboelen:
Thanks for your contribution. It has been included!

PR #720, opened 2019-05-23
Added certificate tests for Exim

dvehrs:
A few more tests to consider for Exim4 in tests_mail_messaging:
  1.  Display where certificates are installed (Default or Custom)
  2.  Show state of "Try Verify Hosts" option
  3.  Show state of "Verify Hosts" option

The last two options may seem slightly confusing but while they have confusing labels they enable different functions.  

When "Try Verify Hosts" is enabled, a system will attempt to verify a hosts certificate but will not close the connection if the verification fails.  This is an appropriate setting for an 'Internet Host' configuration for Exim.

When "Verify Hosts" is enabled, the system attempts to verify a host certificate and will close the connection if verification fails.  This is often an appropriate setting for 'Satellite' or 'Smarthost' configurations.  

If 'Verify Hosts' is enabled, it appears to override 'Try Verify Hosts' so it is possible to see both enabled.

I've tried to reflect the difference of which setting is less desirable by using a YELLOW colored "ENABLED".  Which setting is colored is dependent upon which configuration type is being tested.

What do you think?  Does this give us something to build on for suggestions and/or hardening points?

### DIFF
--- a/include/tests_mail_messaging
+++ b/include/tests_mail_messaging
@@ -161,6 +161,83 @@
                 LogText "Result: Private Key not set."
                 Display --indent 4 --text "- Private Key not set" --result "${STATUS_WARNING}" --color WHITE
             fi
+
+            LogText "Test: Exim Verify Certificates"
+
+            case "${EXIM_TYPE}" in
+                "INTERNET HOST" | "SMARTHOST" )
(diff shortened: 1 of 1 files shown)


mboelen:
Hi @dvehrs - Thanks for your pull request, it has been merged. I replaced the direct calls to exim/cut/sed/grep. If you have the time, have a look.

PR #722, opened 2019-06-06
Detect BSD-style (rc.d) init in Linux systems

pyllyukko:
At least Slackware Linux uses this. Even though it's compatible with SysV init scripts, it's still technically BSD-style init.

I wasn't sure whether to name it `bsdrc` or `bsdrc.d`, but there's no monolithic `/etc/rc` script I think the `rc.d` is more appropriate. Also for instance [this article](https://www.freebsd.org/doc/en/articles/rc-scripting/article.html) uses `rc.d` name for it.

### DIFF
--- a/include/tests_boot_services
+++ b/include/tests_boot_services
@@ -96,7 +96,11 @@
                                 ;;
 
                                 "init" | "initsplash")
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, merged!

PR #727, opened 2019-07-13
Prepend sed to discard '## ' in CHANGELOG.md

Marzal:
Since version 2.6.6 CHANGELOG.md has markdown formatting. This breaks the show changelog command.
This is a workaound to "fix" the issue without changing all the versions strings in CHANGELOG.md

### DIFF
--- a/include/helper_show
+++ b/include/helper_show
@@ -160,15 +160,15 @@ if [ $# -gt 0 ]; then
                 if [ ! -z "${CHANGELOG}" ]; then LogText "Result: found changelog file: ${CHANGELOG}"; break; fi
             done
             if [ ! -z "${CHANGELOG}" ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
Great suggestion and good fix. Thanks for your pull request it has been merged!

Marzal:
Thanks for all the hard work, and for the quick merge.

mboelen:
My pleasure! As you are already familiar with some of the code, do you want to help with testing this new 3.x major release before it goes live?

How you can help: simply git pull so you have the latest version and run as many different commands and combinations as you can (like the 'lynis show changelog' you already used).

Marzal:
OK, I would give it a try. Do you want a report? or should I just open a bug or a PR if I found something?

mboelen:
Thanks! Opening an issue here on GitHub would be great

PR #729, opened 2019-07-17
strictmode fixes

Marzal:
A few changes in order to strict mode to works.
So --profile developer.prf works again too.


mboelen:
Thanks! It is merged

PR #730, opened 2019-07-17
Fix $privileged var being overwrite

Marzal:
With the new reorganization https://github.com/CISOfy/lynis/commit/34f306eb0bc5502535ca85f5700fa083ab6c9443 `. ${INCLUDEDIR}/consts` is called after $PRIVILEGED is set and overwrites the value.
This PR delete the var from that file and set the var at the beginning in both cases


mboelen:
Thanks, merged!

Issue #615, opened 2018-12-08
[CRYPT] or [FILE] check for encrypted partitions


Capashenn:
**Is your feature request related to a problem? Please describe.**
It should be nice if lynis could check for encrypted partitions (LUKS)

**Describe the solution you'd like**
Check for encrypted (LUKS) partitions

**Required changes**
Add a new test in [CRYPT] or [FILE] to check for encrypted (LUKS) partitions
```
$ blkid /dev/sdb1
/dev/sdb1: UUID="xxxxxxxxxxxx" TYPE="crypto_LUKS"   #encrypted
```

pyllyukko:
I think a better test would be:

```
cryptsetup isLuks /dev/sdb1
```

It returns `0` if it's LUKS and `1` when it's not.

mboelen:
Test CRYP-7930 was already implemented to check this. Closing this issue.

PR #731, opened 2019-07-17
[CRYP-7930] Modify to use 'lsblk' and 'cryptsetup'

chr0mag:
There are several challenges with the existing method of using
/etc/crypttab:

1)encrypted rootfs partitions are not typically listed in this
file (users are prompted for password in early boot instead)

2)the 'luks' option is the default option so it is possible for
/etc/crypttab entries to never have this set explicitly and any
block device configured as such is currently missed

3)any device mounted manually, or using any other mechanism aside
from /etc/crypttab will be missed

This commit executes 'cryptsetup isLuks' on every block device in
the system to determine whether it is a LUKS device. This handles
all 3 cases mentioned above and closes issue #615.

Test case wording was also updated to reflect the fact that it
only checks for LUKS encrypted block devices. So, plain dm-crypt
and TrueCrypt/VeraCrypt block device encryption is not detected.
Nor is any file system level encryption such as eCryptfs, EncFs,
gocryptfs.


mboelen:
Thanks, it has been merged.

PR #734, opened 2019-07-19
add Ubuntu nologin path

konstruktoid:
Exclude the default Ubuntu `nologin` shell path so that `HOME-9304` and `HOME-9306` doesn't trigger for e.g. the `daemon` user with the `/usr/sbin` home directory and the `nologin` shell.

### DIFF
--- a/include/tests_homedirs
+++ b/include/tests_homedirs
@@ -57,7 +57,7 @@
     if [ ${SKIPTEST} -eq 0 ]; then
         # Check if users' home directories permissions are 750 or more restrictive
         FOUND=0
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks!

I accidentally fixed the same issue in the meantime using another way: looking at the last part of the string (so everything ending with /false or /nologin). Related commit is https://github.com/CISOfy/lynis/commit/3af72e41c7475821fdc1dcc8e1c90a2cd7fc2a4a

Will close this pull request. Let me know if you feel my patch is worse than this one :)

PR #736, opened 2019-07-29
Document --usecwd in helper_show and man

Marzal:


mboelen:
Thanks, it is merged!

PR #739, opened 2019-08-01
Clean up of not used legacy functions and not used broken functions

Marzal:
For a 3.0 release I think it's ok to clean deprecated and not used code.

None of them are used and some TestCase_* functions are broken, for instance:
`if [ TestCase_Equal "${1}" "${2}" ] || [ TestCase_GreaterThan "${1}" "${2}" ]; then`
and `${FUNCNAME}` works with bash but is not POSIX compliant


mboelen:
Thanks, merged! Will only keep the last part as-is, as this helps people who created custom tests/plugins from getting hard errors and get a software and more clear error message instead.

PR #740, opened 2019-08-01
Fix --user in isRunning allowing word split and cleanup ShellCheck warns

Marzal:
https://github.com/koalaman/shellcheck/wiki/SC2066

Add TODO reminder to test --user functionality on AIX with strict mode

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -1404,17 +1404,18 @@
         if [ -z "${search}" ]; then ExitFatal "Missing process to search for when using IsRunning function"; fi
         RUNNING=0
         # AIX does not fully support pgrep options, so using ps instead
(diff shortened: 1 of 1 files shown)


Marzal:
Hi, do you prefer the patch without the ShellChek directives?

mboelen:
Hi @Marzal - yes, that would be great. The pull request was actually waiting for that, as we are discussing on how to actually do more with ShellCheck, without adding it to the source.

We want to extend the [SDK](https://github.com/CISOfy/lynis-sdk) to handle this. Can you send me a short email (see profile)? I want to discuss a few things with you based on your good pull requests lately, including the usage of ShellCheck :)

mboelen:
Will keep this pull request open, until the other one arrives

Issue #743, opened 2019-08-12
Lynis reports HOME-9304 and HOME-9306 failures for the git user


infokiller:
**Describe the bug**
Lynis reports failures for the tests `HOME-9304` and `HOME-9306` for the git user.

**Version**
 - Distribution [e.g. Ubuntu 18.04]: Archlinux
 - Lynis version [e.g. 2.7.0]: `2.7.5-185-g3006b8d` (commit 3006b8dd267a827af1969099134a443cacf61e7a)

**Expected behavior**
No error reported.

**Output**
If applicable, add output that you get from the tool or the related section of lynis.log

```
2019-08-11 18:14:55 Performing test ID HOME-9304 (Check if users' home directories permissions are 750 or more restrictive)
2019-08-11 18:14:55 Result: permissions of home directory / of user git are not strict enough. Should be 750 or more restrictive. Change with: chmod 750 /
2019-08-11 18:14:55 Warning: Permissions of some users' home directories are not strict enough. Should be 750 or more restrictive. [test:HOME-9304] [details:-] [solution:-]
  2019-08-11 18:14:55 ===---------------------------------------------------------------===
2019-08-11 18:14:55 Performing test ID HOME-9306 (Check if users own their home directories)
2019-08-11 18:14:55 Result: The home directory / of user git is owned by root. Change with: chown git /
2019-08-11 18:14:55 Warning: Owner of some users' home directories are not correctly set [test:HOME-9306] [details:-] [solution:-]
```

**Additional context**
Looking at https://github.com/CISOfy/lynis/blob/3006b8dd267a827af1969099134a443cacf61e7a/include/tests_homedirs#L60 and https://github.com/CISOfy/lynis/blob/3006b8dd267a827af1969099134a443cacf61e7a/include/tests_homedirs#L92 it looks like users with their login shells set to `false` or `nologin` are ignored. Perhaps `git-shell` should also be added?

mboelen:
Sounds like a good improvement. Thanks for reporting!

Implemented with commit https://github.com/CISOfy/lynis/commit/368989d5c4678f662a127de3a078af67df2dbd39 and https://github.com/CISOfy/lynis/commit/2af2d0a89591a52798a9230b4a56e8a47201b500

PR #745, opened 2019-08-15
Fix typos from the fixes for #743

infokiller:
Define 'git' instead of 'git-shell'.


mboelen:
Thanks, it has been merged!

Issue #168, opened 2016-04-16
[HTTP-6710]  SSL nginx checks don't follow includes


tuxayo:
Leading to false positives

mboelen:
Can you show more details? (OS, related directories)

When possible, we appreciate any patch to improve the test.

tuxayo:
OS: Arch Linux

File and directories:
/etc/nginx/nginx.conf

If the ssl conf extracted to another file and then included using `include ssl-example-domain-1.conf;`

In my case there are even two inclusion level:

`/etc/nginx/nginx.conf` includes `example-domain-1.conf` and `example-domain-2.conf`
`example-domain-1.conf` includes `ssl-example-domain-1.conf`
`example-domain-2.conf` includes `ssl-example-domain-2.conf`

vivianspencer:
I just ran into this issue also.

If I include all of my configuration in the host file (/etc/nginx/sites-available/domain.conf) there are no issues. If I separate my SSL parameters into a separate file (/etc/nginx/snippets/ssl-params.conf) I receive a false positive on [HTTP-6710](https://cisofy.com/controls/HTTP-6710/)

OS: Debian 8.5

mboelen:
Hi @tuxayo and @vivianspencer. A rewrite of nginx has been done, to allow parsing of nginx files more extensively. Also included is a second level of file includes. Does it work properly now for you both? The configuration files parsed are displayed on screen (except mime.types).

mboelen:
Any of you two who has a few minutes to test if things are working now? Then we can finish it up for next release.

vivianspencer:
Sorry this is still not working for me, I've checked the log and it's finding the includes but I get the "Result: this include does not point to a file"

mboelen:
Can you share the related include line?

vivianspencer:
Yes sorry, I have several config files in the folder /etc/nginx/snippets,  so the includes come up as "snippets/ssl-params.conf" for instance.

In my log this is what I get:

>
2016-08-19 07:46:49 Result: found option include with parameters snippets/cache.conf
2016-08-19 07:46:49 Found unknown option include in nginx configuration
2016-08-19 07:46:49 Result: found option include with parameters snippets/drupal.conf
2016-08-19 07:46:49 Found unknown option include in nginx configuration
2016-08-19 07:46:49 Result: found option include with parameters snippets/restrictions.conf
2016-08-19 07:46:49 Found unknown option include in nginx configuration
[...]

PR #761, opened 2019-09-18
enhance nginx include config detection …

Schmuuu:
... to correctly include configs via relative path and wildcards

Hello,
These changes should fix https://github.com/CISOfy/lynis/issues/168
I had a similar issue and noticed that the root cause of my and this issue are the missing support for
- relative paths
- wildcard includes (like sites-enabled/*.conf)

In both cases the conf files were not found and therefor not parsed.
I included the necessary code, which works perfectly fine on 
- bash 5.0.9
- ksh 2020.0

Unfortunately I have no clue how to install bourne-shell on Arch Linux to test that as well :(

This is my first real contribution with bigger code-changes to an open-source project, so please have a little patience with me. I'm willing to do further code changes, if you could show me the direction how to better test or what to change to support bourne-shell as well. Thanks a lot!


mboelen:
Hi @Schmuuu - Thanks for this first contribution. Looks good to me, so it is merged.

There is a helpful page regarding testing against /bin/sh for Arch. See the [dash page](https://wiki.archlinux.org/index.php/Dash) on the Arch wiki.

PR #765, opened 2019-09-19
Fix auth-9282

### LINKED ISSUE #608: AUTH-9282 : Search password protected accounts without expire date

Marzal:
This fixes #608.

`Register function` uses/overwrite variable `FIND` in 
https://github.com/CISOfy/lynis/blob/master/include/functions#L2571
so it needs to be registered before
https://github.com/CISOfy/lynis/blob/master/include/tests_authentication#L740
or use another variable like in this PR

Also I think there is a typo in 

> found one or more accounts with**out** expire date set

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -742,24 +742,24 @@
             case ${LINUX_VERSION} in
                 "SuSE")
                     PREQS_MET="YES"
-                    FIND=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="P" && $5=="99999") print $1 }')
+                    FIND_P=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="P" && $5=="99999") print $1 }')
                     FIND2=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="NP") print $1 }')
                     ;;
                 *)
                     PREQS_MET="YES"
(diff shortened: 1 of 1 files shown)


Marzal:
I did further testing, moving the Register function [up](https://github.com/CISOfy/lynis/blob/master/include/tests_authentication#L740) don't work. The 2 options that I see are changing the name of the variable inside the prerequisite test (this patch) or in the `Register function` itself.
https://github.com/CISOfy/lynis/blob/master/include/functions#L2571
The second approach has the advantage that fixes bugs similar to #608

mboelen:
Looks good, merged!

PR #766, opened 2019-09-20
Update tests_ldap

multiwebinc:
Check was backwards


mboelen:
Thanks, good catch. Merged!

PR #770, opened 2019-09-26
nginx-test: add log file detection for relative paths

Schmuuu:
Hi again :)

Just like in https://github.com/CISOfy/lynis/pull/761 I added support for relative paths. So whenever someone uses something like

```
error_log  logs/error.log;
access_log  logs/access.log;
```
these log files are found now.

Tested with:
- bash 5.0.9
- dash 0.5.10.2

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -2181,8 +2181,9 @@
                         if [ ! "${VALUE}" = "" ]; then
                             # If multiple values follow, select first one
                             VALUE=$(echo ${VALUE} | awk '{ print $1 }')
-                            if [ ! -f ${VALUE} ]; then
-                                LogText "Result: could not find referenced log file ${VALUE} in nginx configuration"
+                            # Find both, log files provided with full or relative path
+                            if [ ! -f ${VALUE} -a ! -f "${CONFIG_FILE%nginx.conf}${VALUE}" ]; then
+                                LogText "Result: could not find log file ${VALUE} referenced in nginx configuration"
                                 NGINX_ACCESS_LOG_MISSING=1
                             fi
                         fi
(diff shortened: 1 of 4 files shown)


Schmuuu:
Oh, didn't know this pull request is automatically updated with additional commits. 
Cool :+1: 
So here are the details for the other commits then...

Commit https://github.com/CISOfy/lynis/pull/770/commits/07084501b6e4eb2c796431a95f2f708ed634c954
Helps to detect LTS kernels on Arch-Linux. 
Before:
```
Exception found!
Function/test:  [KRNL-5830:3]
Message:        Could not find our running kernel on disk, which is unexpected
...
- Check if reboot is needed                                 [ UNKNOWN ]
```
Afterwards:
```
- Check if reboot is needed                                 [ NO ]
```

----------------------

Commit https://github.com/CISOfy/lynis/pull/770/commits/b38031803d1350aa6af89663631fa7abe833219a
Avoids grep error outputs in case the added folder from https://github.com/CISOfy/lynis/commit/1df9630bcfcf1134cb1f4323cf12d0229181d929 is missing.
Before:
```
- Checking Linux kernel configuration file                  [ FOUND ]
/usr/bin/grep: /etc/security/limits.d: No such file or directory
/usr/bin/grep: /etc/security/limits.d: No such file or directory
- Checking core dumps configuration                         [ NOT FOUND ]
```
After:
```
- Checking Linux kernel configuration file                  [ FOUND ]
- Checking core dumps configuration                         [ NOT FOUND ]
``` 
(Not found is so far correct. The test needs some more changes to run with my setup. I'll probably add the necessary code lines later)

---------------

All tested with bash and dash again.

mboelen:
Thanks for all improvements! Next please create separate pull requests :)

PR #772, opened 2019-09-29
Clean shellcheck warnings in STRG-1846

Marzal:
Also avoid false positive in lynis-sdk, plus the usual double quote to prevent globbing and word splitting


Schmuuu:
Hi,

Nice to see your pull request. 
May I ask: This (`if [ -n "${FIND1}" ] || [ -n "${FIND2}" ]; then`) is definitely the way to go then?

When I wrote additional code I was looking around in the modules and was trying to write my code just like it was done before and therefor I was using the -o variant. I wasn't sure why this was preferred before so I just kept using it. 

If double quotes and the || or && should definitely be used in the project, I would change my code accordingly, but wanted to have confirmation first.

Thanks a lot!

mboelen:
Thanks, merged!

PR #777, opened 2019-10-08
Slackware Linux OS detection

pyllyukko:
What kind of tool is this if it would fail to detect the oldest living distro?! :smile: 

The `os-release` can be seen here: <ftp://ftp.slackware.com/pub/slackware/slackware64-current/source/a/aaa_base/os-release>


mboelen:
Thanks!

PR #779, opened 2019-10-08
Redirect gradm stderr to /dev/null

pyllyukko:
If the tool exists but the kernel doesn't have grsec, you'll get the
following error into stderr:

Could not open /dev/grsec.
open: No such file or directory


mboelen:
Thanks!

Issue #701, opened 2019-03-26
Direct root Logins Not Allowed


Capashenn:
**Is your feature request related to a problem? Please describe.**
Disabling direct root logins ensures proper accountability and multifactor authentication to privileged accounts. Users will first login, then escalate to privileged (root) access via su / sudo. This is required for FISMA Low and FISMA Moderate systems.

**Describe the solution you'd like**
Check if /etc/securetty is empty

**Required changes**
Add a new test

pyllyukko:
>  Check if /etc/securetty is empty

To make sure, it would be also required to check whether [pam_securetty](http://www.linux-pam.org/Linux-PAM-html/sag-pam_securetty.html) (which is in charge of enforcing `/etc/securetty`) is used by services utilizing PAM.

PR #785, opened 2019-10-14
[AUTH-9411] Test for /etc/securetty

katkad:
fixes https://github.com/CISOfy/lynis/issues/701

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -1480,6 +1480,32 @@
 #
 #################################################################################
 #
+    # Test        : AUTH-9411
+    # Description : Check if /etc/securetty is empty
+    Register --test-no AUTH-9411 --weight L --network NO --category security --description "Checking /etc/securetty file"
+    if [ ${SKIPTEST} -eq 0 ]; then
+        LogText "Test: checking /etc/securetty"
+        if [ -f /etc/securetty ]; then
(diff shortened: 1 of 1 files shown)


include/tests_authentication
```diff
@@ -1479,6 +1479,31 @@
...
 #################################################################################
+#
+    # Test        : AUTH-9411
+    # Description : Check if /etc/securetty is empty
+    Register --test-no AUTH-9411 --weight L --network NO --category security --description "Checking /etc/securetty file"
+    if [ ${SKIPTEST} -eq 0 ]; then
+        LogText "Test: checking /etc/securetty"
+        if [ -f /etc/securetty ]; then
+            LogText "Result: file /etc/securetty found"
+            FIND=$(${GREPBINARY} -v "^#" '/etc/securetty' | wc -l)
```
Schmuuu: This count empty lines as well and with only one empty line, this test will always complain. 
It's probably better to use:

```
# use of "^ *" instead of "^\s*", because HPUX und Solaris can't handle "\s". Using " *" seems work on all platforms though
FIND=$(${GREPBINARY} -v "^ *#" '/etc/securetty' | grep -v "^ *$ | wc -l)
```

katkad: Thank you for review. I'll fix it.

PR #792, opened 2019-10-16
Correctly split lines of /etc/passwd

### LINKED ISSUE #773: debian / false positive for some users on  HOME-9306
**Describe the bug**
Some users are reported having a homedirectory which is not owned by them. This is wrong.

**Version**
 - Debian 10.1
 - Lynis version master (downloaded from github as zip)

**Expected behavior**
Get no error on such users

**Output**

**Additional context**
for example, my amavis user looks like: 
`amavis:x:114:120:AMaViS system user,,,:/var/lib/amavis:/bin/sh
`
The report says: 
`2019-10-06 21:26:11 Result: The home directory  of user amavis is owned by root. Change with: chown amavis`

But when i check the folder /var/lib/amavis, it says:
`drwxr-x---  7 amavis         amavis        4,0K Okt  6 21:17 amavis`

Aresius423:
Fixes #773

```
# cat /etc/passwd

ec2-user:x:501:501:EC2 Default User:/home/ec2-user:/bin/bash


# for LINE in $(grep -v '^(daemon|git|halt|root|shutdown|sync)' /etc/passwd | awk -F: '($7 !~ "/(false|nologin)$") { print }'); do echo $LINE; done  

[...]  
ec2-user:x:501:501:EC2  
Default  
User:/home/ec2-user:/bin/bash  
```

### DIFF
--- a/include/tests_homedirs
+++ b/include/tests_homedirs
@@ -57,7 +57,8 @@
     if [ ${SKIPTEST} -eq 0 ]; then
         # Check if users' home directories permissions are 750 or more restrictive
         FOUND=0
-        for LINE in $(${EGREPBINARY} -v '^(daemon|git|halt|root|shutdown|sync)' ${ROOTDIR}etc/passwd | ${AWKBINARY} -F: '($7 !~ "/(false|nologin)$") { print }'); do
+        USERDATA=$(${EGREPBINARY} -v '^(daemon|git|halt|root|shutdown|sync)' ${ROOTDIR}etc/passwd | ${AWKBINARY} -F: '($7 !~ "/(false|nologin)$") { print }')
+	while read -r LINE; do
             USER=$(echo ${LINE} | ${CUTBINARY} -d: -f1)
             DIR=$(echo ${LINE} | ${CUTBINARY} -d: -f6)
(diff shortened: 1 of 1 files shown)


neilmayhew:
This doesn't work for me. I get:

```
$HOME/src/lynis/include/functions: line 553: echo -e: command not found
```

on several tests.

However, the fix given by @church1e in his comments on #773 does work.

Aresius423:
@neilmayhew you're right - with some shells, the parameter expansion I used in the previous commit doesn't work, and IFS not being reset breaks the Display function, hence the error message.

Could you please check it in your environment with the new commit?

neilmayhew:
Unfortunately, that doesn't work. It fails to pick up a true positive when I set one of the home directory's permissions to 755. I think the problem is that subshells are incompatible with the design of lynis, which sets variables inside loops and then checks them outside. In this case, it's the `FOUND` variable.

Aresius423:
Third time's the charm. I haven't found a prettier (but still portable) way to do it, but at least this avoids having to deal with IFS altogether.

mboelen:
Thanks, the change has been merged! Let's see if this resolves it.

Thanks @neilmayhew for your guidance in this issue.

PR #795, opened 2019-10-18
fix for #781 - run lsblk without --paths

### LINKED ISSUE #781: RHEL 6 - /bin/lsblk: unrecognized option '--paths' doing Cryptography checks
**Describe the bug**
Running Lynis on RHEL 6.10 I see under "[+] Cryptography" the following:

```
[+] Cryptography
- Checking for expired SSL certificates [0/23] [ NONE ]
/bin/lsblk: unrecognized option '--paths'

Usage:
 lsblk [options] [<device> ...]

Options:
 -a, --all            print all devices
 -b, --bytes          print SIZE in bytes rather than in human readable format
 -d, --nodeps         don't print slaves or holders
 -D, --discard        print discard capabilities
 -e, --exclude <list> exclude devices by major number (default: RAM disks)
 -I, --include <list> show only devices with specified major numbers
 -f, --fs             output info about filesystems
 -h, --help           usage information (this)
 -i, --ascii          use ascii characters only
 -m, --perms          output info about permissions
 -l, --list           use list format ouput
 -n, --noheadings     don't print headings
 -o, --output <list>  output columns
 -P, --pairs          use key="value" output format
 -r, --raw            use raw output format
 -s, --inverse        inverse dependencies
 -t, --topology       output info about topology
 -V, --version        output version information and exit

Available columns (for --output):
        NAME  device name
       KNAME  internal kernel device name
     MAJ:MIN  major:minor device number
      FSTYPE  filesystem type
  MOUNTPOINT  where the device is mounted
       LABEL  filesystem LABEL
        UUID  filesystem UUID
          RA  read-ahead of the device
          RO  read-only device
          RM  removable device
       MODEL  device identifier
        SIZE  size of the device
       STATE  state of the device
       OWNER  user name
       GROUP  group name
        MODE  device node permissions
   ALIGNMENT  alignment offset
      MIN-IO  minimum I/O size
      OPT-IO  optimal I/O size
     PHY-SEC  physical sector size
     LOG-SEC  logical sector size
        ROTA  rotational device
       SCHED  I/O scheduler name
     RQ-SIZE  request queue size
        TYPE  device type
    DISC-ALN  discard alignment offset
   DISC-GRAN  discard granularity
    DISC-MAX  discard max bytes
   DISC-ZERO  discard zeroes data

For more details see lsblk(8).
- Kernel entropy is sufficient [ NO ]
```
image

**Version**
Distributor ID:	RedHatEnterpriseServer
Description:	Red Hat Enterprise Linux Server release 6.10 (Santiago)
Release:	6.10
Codename:	Santiago
Lynis: 3.0.0

**Expected behavior**
I'm not sure what the Cryptography check is using lsblk for as I'm not seeing this in the logs. The RHEL 6 version is probably too old.

Schmuuu:
Hi,

This should fix #781 and allow compatibility with older Linux systems where lsblk doesn't support --paths yet.
Tested with bash and dash on Manjaro and Arch.

I'm going to post this pull request already and could run a test on RedHat 6 next week Mo or Tue if requested.

### DIFF
--- a/include/tests_crypto
+++ b/include/tests_crypto
@@ -138,9 +138,9 @@
         FOUND=0
 
         # cryptsetup only works as root
-        if [ -n "${LSBLKBINARY}" -a -n "${CRYPTSETUPBINARY}" -a ${FORENSICS_MODE} -eq 0 ]; then
-            for BLOCK_DEV in $(${LSBLKBINARY} --noheadings --list --paths -o NAME); do
-                if ${CRYPTSETUPBINARY} isLuks ${BLOCK_DEV} 2> /dev/null; then
+        if [ -n "${LSBLKBINARY}" ] && [ -n "${CRYPTSETUPBINARY}" ] && [ ${FORENSICS_MODE} -eq 0 ]; then
+            for BLOCK_DEV in $(${LSBLKBINARY} --noheadings --list -o NAME 2> /dev/null | cut -d' ' -f1); do
+                if ${CRYPTSETUPBINARY} isLuks $(${FINDBINARY} /dev/ -name "${BLOCK_DEV}" 2> /dev/null) 2> /dev/null; then
                     LogText "Result: Found LUKS encrypted block device: ${BLOCK_DEV}"
                     Report "encryption[]=luks,block_device,${BLOCK_DEV}"
                     FOUND=$((FOUND +1))


mboelen:
Thanks for the pull request and testing work with @daniejstriata on this. It has been merged!

PR #796, opened 2019-10-19
Additional test for permissions on sudoers files

dvehrs:
Noticed that the tests for sudo config files were only testing three acceptable permission combinations.

It tested for one of:
   rw-------
   rw-rw----
   r--r-----

But there was one option missing that I happen to use:  rw-r-----

So I added it in a simple patch.  What do you think?  Or should we also add a condition for only the owner being able to read the file (r--------)?

Dave

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -513,7 +513,7 @@
             FIND=$(${LSBINARY} -l ${f} | ${CUTBINARY} -c 2-10)
             FIND2=$(${LSBINARY} -n ${f} | ${AWKBINARY} '{print $3$4}')
             LogText "Result: Found file permissions: ${FIND} and owner UID GID: ${FIND2}"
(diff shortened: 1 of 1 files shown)


dvehrs:
I've been comparing the permissions checks in FILE-7524 and AUTH-9252.  I've noticed a significant difference.  AUTH-9252 (and AUTH-9410) verify correct permissions by comparing a segment of the string output of the ls command with other known good strings.  However FILE-7524 uses the HasCorrectFIlePermissions function which uses a couple methods to get the octal permissions for the file then does a numeric comparison.  

I wondering if it wouldn't be better for all to use a numeric comparison rather than simply adding more strings to the existing comparisons.

Is this a road worth going down or am I missing a valid reason for why the comparisons are different?

dvehrs:
I think I might have found a better solution for this.  I am testing a few ideas, if it works out I will submit that change.  If it doesn't work out I will resubmit this one.

PR #800, opened 2019-10-22
don't fail relative paths check with spaces in PATH

### LINKED ISSUE #768: Incorrect report of relative path in PATH for paths with spaces
**Describe the bug**

I get `Found relative path in PATH` when there is no relative path in it. Having a path with spaces, like `/home/niklas/opt/sublime/Sublime Text 2`, in `PATH`, creates this false positive.

**Version**
 - Ubuntu 16.04
 - Lynis version: 3.0.0 (commit 0088593506138b54f6edcabf995fdaac5cbeee9a)

**Expected behavior**

No warning.

**Output**

```
  Warnings (1):
  ----------------------------
  ! Possible riskful location in PATH discovered [CORE-1000] 
    - Details  : text:Found relative path in PATH
      https://cisofy.com/lynis/controls/CORE-1000/
```

**Additional context**

My colleague @neilmayhew pointed out:

> I ran shellcheck on the Lynis source. It came up with a LOT of Double quote to prevent globbing and word splitting warnings. Interestingly, there was an issue (now closed) to use shellcheck and fix as many issues as possible (https://github.com/CISOfy/lynis/issues/189).
> The author didn't fix this one, though:
> `PATH_REVERSED=$(echo ${PATH} | awk -F: '{ for (i=NF; i>1; i--) printf("%s ",$i); print $1; }')`

Schmuuu:
This could fix #768
I'll request a test in the issue. With my tests everything was alright, but could be different in a real world scenario.

Anyway: tested with bash and dash again.

### DIFF
--- a/include/binaries
+++ b/include/binaries
@@ -50,7 +50,7 @@
         # Test if our PATH variable provides a set of paths. If so, reverse the order. If we discover the same binary
         # multiple times, the one first in PATH should be used.
         if [ -n "${PATH}" ]; then
-            PATH_REVERSED=$(echo ${PATH} | awk -F: '{ for (i=NF; i>1; i--) printf("%s ",$i); print $1; }')
+            PATH_REVERSED=$(echo "${PATH}" | sed 's/ /!!space!!/g' | awk -F: '{ for (i=NF; i>1; i--) printf("%s ",$i); print $1; }')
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, merged!

PR #803, opened 2019-10-23
Updated tests for file permissions to case statements

dvehrs:
Converted file permission tests from if statements to case statements.  This allows us to forgo the need to a chain of 'and/or' conditions and use some basic regular expressions.  Formatted the sudoers file tests so that the progress display looks the same but the log file will give the user a little bit more information on what needs to be updated.

This method seems easier to read to me, what do you think?

Dave


mboelen:
Not sure yet about the readability part. It is shorter in code, but we have to find out if it is maintainable. As I want to give it a try, I have merged it. Thanks! :)

PR #804, opened 2019-10-23
fix left over test folder with correct folder from variable

### LINKED ISSUE #801: KRNL-5820 - incorrect path
**Describe the bug**
Test **KRNL-5820 in tests_kernel** partly sets /etc/security/limits.conf.d as path to check for files, in the example of a CentOS system the path is /etc/security/limits.d

This leads to some settings not being detected although they are correctly set, funnily enough in the log the correct path is mentioned.

**Version**
 - Distribution CentOS 8.0.1905
 - Lynis version 3.0.0

**Expected behavior**
The correct paths should be detected on all systems.

**Output**
**2019-10-23 15:22:24 Test: Checking if core dumps are disabled in /etc/security/limits.conf and /etc/security/limits.d/*
2019-10-23 15:22:24 Result: core dumps are not explicitly disabled
2019-10-23 15:22:24 Suggestion: If not required, consider explicit disabling of core dump in /etc/security/limits.conf file [test:KRNL-5820] [details:-] [solution:-]
2019-10-23 15:22:24 Hardening: assigned partial number of hardening points (1 of 3). Currently having 9 points (out of 11)**

**Additional context**
The file /etc/security/limits.d/99-coredump.conf contains the following:
_# Disable core dumps for all users_
_* soft core 0_
_* hard core 0_

But the test doesn't recognize it.

Schmuuu:
When adding the feature to test additional locations for disabled coredumps I left over a test folder.
Fixed that now by using the variable which was already there, which I only forgot to replace the folder with.
Fixes #801


mboelen:
Merged!

PR #809, opened 2019-10-30
First fix in order make lynis work OK in AIX

Marzal:
grep doesn't work with '\|' in AIX. grep -E is more POSIX compliant.

Without this fix, lynis doesn't work at all


mboelen:
Thanks, merged!

PR #810, opened 2019-11-01
Fix for false positive for some users on HOME-9306

### LINKED ISSUE #773: debian / false positive for some users on  HOME-9306

neilmayhew:
Closes #773

Patch taken from issue comments by @church1e

---
This change is [Reviewable](https://reviewable.io/reviews/cisofy/lynis/810)

### DIFF
--- a/include/tests_homedirs
+++ b/include/tests_homedirs
@@ -57,9 +57,9 @@
     if [ ${SKIPTEST} -eq 0 ]; then
         # Check if users' home directories permissions are 750 or more restrictive
         FOUND=0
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks Neil, looking good and merged!

Issue #793, opened 2019-10-16
Crypto: Check for TRNG and rngd


pyllyukko:
**Is your feature request related to a problem? Please describe.**
Lynis could check for existence of TRNG and whether `rngd` is running.

**Describe the solution you'd like**
* Check that `/sys/class/misc/hw_random/rng_current` value is not "none"
* Check that `rngd` is running and utilizing the TRNG

**Required changes**
* [x] Assign `CRYP` ID (CRYP-8004)
* [ ] Assign potential hardening points
* [ ] Implement a test for this

**Additional context**
Somewhat relates to #751

mboelen:
Sounds like a good change. Want to submit an initial pull request for this?

Suggested CRYP id = CRYP-8004

pyllyukko:
> Sounds like a good change. Want to submit an initial pull request for this?

Sure.

mboelen:
Implemented with pull request #821

PR #821, opened 2020-01-06
Added CRYP-8004

pyllyukko:
Initial version of CRYP-8004 (see #793).

There are few shortcomings:

* This will not detect all RNGs (for instance OneRNG registers itself as `/dev/ttyACM0` and not through Linux's HW RNG drivers)
* There is no verification that `rngd` is actually using the found HW RNG. It could as well be using `/dev/zero` etc. This can be implemented, but might require a bit more complex check, as the input can be configured in several places (conf files, cmdline parameters etc.).

### DIFF
--- a/include/tests_crypto
+++ b/include/tests_crypto
@@ -195,6 +195,33 @@
 #
 #################################################################################
 #
+    # Test        : CRYP-8004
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, merged!

PR #828, opened 2020-01-22
FILE-6310 for HP-UX

gfelkel:
HP-UX: /usr/sbin/mount reports "/home on /dev/…", so $1 has to be used


mboelen:
Thanks for the feedback and pull request. It has been merged!

PR #829, opened 2020-01-23
AUTH-9228 for HP-UX

gfelkel:
HP-UX also has /usr/sbin/pwck. For trusted systems, two additional options -s (check inconsistencies with the protected password database) and -l (check encrypted password lengths that are greater than 8 characters) are available.


mboelen:
Thanks, merged!

PR #833, opened 2020-01-28
NETW-3014: Report correct promisc interface

aperum:
NETW-3014 had its iterator name changed from ${I} to ${ITEM} 3 years ago but the report lines still reference ${I}. This reports the last thing ${I} was set to instead of the actual interface.


mboelen:
Thanks, good catch! The change has been merged.

Issue #528, opened 2018-03-06
Remove test for deprecated OpenSSH config option VerifyReverseMapping


artiommocrenco:
`VerifyReverseMapping` OpenSSH config option is deprecated.

```
$ sshd -t
`/etc/ssh/sshd_config line x: Deprecated option VerifyReverseMapping`
```

I suggest removing it from OpenSSH configuration tests.

mboelen:
Thanks for the suggestion. Although it may be deprecated, older OpenSSH versions might still be using it.

If you remove the particular line, do you get a warning or suggestion?

mboelen:
@artiommocrenco ?

mboelen:
Closing issue for now.

kebertxela:
```shell=
user@host~ sudo grep Verify /etc/ssh/sshd_config 
VerifyReverseMapping yes
user@host~ sudo sshd -t
/etc/ssh/sshd_config line 31: Deprecated option VerifyReverseMapping
user@host~ sudo sed -i 's/Verify/\#Verify/' /etc/ssh/sshd_config 
user@host~ sudo sshd -t
user@host~ 
```

NicolasCARPi:
I have submitted a PR to address this issue: https://github.com/CISOfy/lynis/pull/840

PR #840, opened 2020-02-18
Remove the test for ssh config VerifyReverseMapping

NicolasCARPi:
This option is deprecated since 2003. Having it in a config file raises
a warning and UseDNS (that is on by default) includes the
VerifyReverseMapping check.

See
https://github.com/openssh/openssh-portable/commit/3a961dc0d36c1f87788b707130f6d07709822d38

See #528

--------
I believe it's safe to remove a check for something deprecated for 17 years ;)

Cheers,
~Nico

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -151,7 +151,6 @@
                 StrictModes:YES,,NO:=\
                 TCPKeepAlive:NO,,YES:=\
                 UseDNS:NO,,YES:=\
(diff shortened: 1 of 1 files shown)


NicolasCARPi:
Hello,

Can someone please have a look and possibly merge this?

Cheers :)

~Nico

mboelen:
Thanks, it has been merged!

PR #852, opened 2020-03-03
Added detection of PureOS in /etc/os-release

craigcomstock:
Add detection of PureOS

### DIFF
--- a/include/osdetection
+++ b/include/osdetection
@@ -180,6 +180,13 @@
                             OS_VERSION_FULL=$(grep "^VERSION=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
                             OS_NAME="Debian"
                         ;;
+                        "pureos")
+                            LINUX_VERSION="PureOS"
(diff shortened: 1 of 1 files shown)


include/osdetection
```diff
@@ -180,6 +180,13 @@
                             OS_VERSION_FULL=$(grep "^VERSION=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
                             OS_NAME="Debian"
                         ;;
+                        "pureos")
+                            LINUX_VERSION="PureOS"
+                            OS_FULLNAME=$(grep "^PRETTY_NAME=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
```
nickanderson: instead of `grep | awk | tr`, how about awk `awk -F= '/^PRETTY_NAME=/ {print substr($2,2,length($2)-2)}' /etc/os-release`?

mboelen: Thanks Nick, that is a good suggestion. Will apply that as a default to avoid repetition.

PR #853, opened 2020-03-04
Skip the PHP cli configuration file when looking for expose_php

### LINKED ISSUE #849: expose_php on but it is off
**Describe the bug**
When i run lynis master it reports that php_expose is On, which is correct for cli but not for the web version (which is actually used).
It is on in the php.ini in /etc/php/7.3/cli/php.ini, but its Off in /etc/php/7.3/apache2/php.ini and /etc/php/7.3/fpm/php.ini

**Version**
 - Debian 10
 - Lynis version master

**Expected behavior**
no error reported 

**Output**

**Additional context**

NicolasCARPi:
The expose_php configuration option is only relevant for non-cli PHP and
thus lynis should not look for it in config files that are for cli

Fix #849

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ Using the relevant options, the scan will change base on the intended goal.
 - NAME-4404 - improved screen and log output
 - NAME-4408 - corrected Report function call
 - NETW-3032 - small rewrite of test and extended with addrwatch
+- PHP-2372 - don't look in the cli configuration files
 - PKGS-7410 - use multiple package managers when available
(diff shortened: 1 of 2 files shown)


mboelen:
Thanks for the pull request. Unfortunately, this is not POSIX compatible. So that should be rewritten (e.g. using grep and "/cli/" is in path). Do you want to rewrite it?

NicolasCARPi:
@mboelen Yes I'll try and provide a POSIX compatible implementation. I'll just push another commit on this branch.

NicolasCARPi:
@mboelen POSIX compliant test committed!

PR #854, opened 2020-03-17
Fix quote syntax error in manpage

duck-rh:
Quotes have a special meaning when used at the start of line.


mboelen:
Thanks, merged!

PR #857, opened 2020-03-19
Check if /vmlinuz is missing due to /etc/kernel-img.conf

topimiettinen:
If /etc/kernel-img.conf has the line do_symlinks=No, Debian (probably
also Ubuntu) kernel packages will not update /vmlinuz
etc. symlinks. In that case, guess the kernel from uname -r.


mboelen:
Thanks for another good suggestion and pull request. It has been merged!

PR #858, opened 2020-03-19
Fix logging of running and enabled services

topimiettinen:
Log lines for running and enabled services were mixed up, fix.


mboelen:
Thanks, good catch! It has been merged.

PR #861, opened 2020-03-19
Enhance SELinux checks

topimiettinen:
Display and log: permissive types (rules are not enforced), unconfined
processes (not confined by rules) and processes with initrc_t
type (generic type with weak rules).


mboelen:
Merged, thanks!

PR #862, opened 2020-03-19
FS module tests: check if modules are blacklisted

topimiettinen:
Check if FS modules are blacklisted, also suggest 'blacklist module'
instead of obsolete syntax.

### DIFF
--- a/include/tests_filesystems
+++ b/include/tests_filesystems
@@ -792,6 +792,12 @@
                     AddHP 3 3
                     if IsDebug; then Display --indent 6 --text "- Module ${FS} not present in the kernel" --result OK --color GREEN; fi
                 fi
+                FIND1=$(${EGREPBINARY} "blacklist ${FS}" ${ROOTDIR}etc/modprobe.d/* | ${GREPBINARY} -v "#")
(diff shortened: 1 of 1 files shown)


mboelen:
This pull request needs a little change. There is a difference between redirecting a module to /bin/true and a blacklist statement. The first will not allow loading modules manually (so a true blacklist), where `blacklist` will only prevent _automatic_ loading.

topimiettinen:
Ugh, that's not good. I'll file a bug with kmod to propose an alternative, since they want to get rid of 'install' rules.

topimiettinen:
Sent a question to the list: https://lore.kernel.org/linux-modules/c38d8c8d-5290-7ff7-8daa-d8d76dc80ce0@gmail.com/T/#u

mboelen:
For now, I suggest to undo the textual change.

Other suggestion: test only for /bin/true when testing for `install`. To my understanding /bin/false returns a negative and might believe the loading process to believe something failed (which in fact you want the 'blacklist action' to succeed and give this signal back by using /bin/true).

topimiettinen:
Updated. I haven't received any response from the list so far, but if suddenly there's a new way for alwasy blacklisting, it can be added later.

mboelen:
Yup, if new insights pop up, it will be easy to add. We just need to keep in mind that new features aren't always available to older operating systems (especially the Long Time Support versions).

This one is merged now!

PR #863, opened 2020-03-19
pam: ignore leading dash

topimiettinen:
PAM rules may have a leading '-' character to indicate that if the
module is missing, the error will be ignored, so let's ignore it in
the check.


mboelen:
Thanks, merged!

PR #866, opened 2020-03-19
Run 'systemd-analyze security'

topimiettinen:
'systemd-analyze security' makes a nice overall evaluation of
hardening levels of services in a system. More details can be found
with 'systemd-analyze security SERVICE' for each service.


mboelen:
Thanks for another good suggestion. There is one 'thing': it does not work on older systemd versions. So it looks like this test needs to check first what systemd version is being used.

topimiettinen:
OK, added test to check systemd version. At least v240 (released 2018-12-21) is needed.

mboelen:
Thanks and especially for the eye for detail to add it also to the change log :)

PR #869, opened 2020-03-20
Add AUTH-9234 for NetBSD

bginsbach:
This adds NetBSD in to the supported OSes for AUTH-9234. It is possible that this could could be shared with OpenBSD. It was kept separate in case OpenBSD does not support a `/etc/usermgmt.conf` file for setting the default `useradd(8)` UID range.

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -331,7 +331,7 @@
     # Notes       : AIX: 100+
     #               HPUX: 100+
     #               macOS doesn't have any user info in /etc/passwd, users are managed with opendirectoryd)
-    #               OpenBSD/NetBSD: unknown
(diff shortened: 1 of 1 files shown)


mboelen:
Looks like OpenBSD supports [/etc/usermgmt.conf](https://man.openbsd.org/usermgmt.conf.5), so I suggest to rewrite this test and simply add it.

Can you change your code to reflect this? (so add NetBSD to same line as OpenBSD and keep in your comment on range)

bginsbach:
Agree. Combined NetBSD and OpenBSD logic so both now use `/etc/usermgmt.conf` to determine default normal user UID range. Branch updated with change.

mboelen:
Merged, thanks!

PR #870, opened 2020-03-20
Make BOOT-5260 Linux only

bginsbach:
Linux is the only OS with systemd so no need to check for systemd
single user mode on other operatings systems.


mboelen:
For now that is true, so change merged. Thanks!

PR #871, opened 2020-03-20
Fix uses of non-standard find not operator

bginsbach:
Use `!` rather than the non-standard `find(1) -not` operator.


mboelen:
Thanks. It is indeed more POSIX-alike although less friendly to read. Pull request is merged!

PR #872, opened 2020-03-20
Add NetBSD EOL

bginsbach:
Slightly modify how the EOL lookup works in osdetection by changing the
semantic meaning of an empty date field (field 3). If this field is empty then
the EOL for the OS is false (0). This is much clearer than setting a date far
in the future. There are some OS that have a rolling upgrade (Arch Linux)
and others that EOL older releases after a certain number of (major) releases
have past (NetBSD).

Add EOL data for NetBSD based on NetBSD's general policy of marking a
release EOL one month after the second succeeding major release.


mboelen:
Thanks, sounds like a better method, so it has been merged.

PR #874, opened 2020-03-21
Check password hashing methods

topimiettinen:
Manual page crypt(5) gives recommendations for choosing password
hashing methods, so let's check if there are weakly encrypted
passwords in the system.


mboelen:
Thanks, merged!

Note: made the test root-only, as it requires access to /etc/shadow.

PR #877, opened 2020-03-22
Add FreeBSD and NetBSD to AUTH-9268

bginsbach:
Add FreeBSD and NetBSD as both support PAM. Simplify the `PREQS_MET`
set test by using a case rather than a long if or.


mboelen:
Thanks, merged!

PR #878, opened 2020-03-22
Check IMA/EVM, dm-integrity and dm-verity statuses

topimiettinen:
Check for evmctl (Extended Verification Module) tool and system IMA (Integrity Measurement
Architecture) status.


topimiettinen:
I was also thinking about checks for integritysetup and veritysetup (which are probably more popular than IMA/EVM), but I don't have an idea how to find out if they are used. They could be installed as part of a bigger package (cryptsetup), so just the presence of the tools does not tell much.

topimiettinen:
Wasn't so hard after all. I put this commit on top of EVM/IMA commit to avoid merge conflicts.

mboelen:
Thanks, merged!

mboelen:
Looks like a file was missing. Made commit https://github.com/CISOfy/lynis/commit/603d5b16a282b6c2f646bae450002e385bd91b3f

@topimiettinen Is it correct this way?

PR #879, opened 2020-03-23
Enhance TOMOYO Linux check

topimiettinen:
Count and log unconfined processes, which are not using policy
profile 3.


mboelen:
Thanks, merged!

PR #882, opened 2020-03-27
[CRYP-7902] Check also certificates provided by packages

topimiettinen:
The package maintainers are not immune to mistakes or they might not
always provide timely updates, so let's check more certificates even
if they are delivered by packages.

I found three expired certificates in my Debian/unstable system,
thanks to changed Lynis.


topimiettinen:
Opened a bug report [955038](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=955038) for Debian package ca-certificates.

mboelen:
The reason to skip testing certificates installed by packages is that some CA certificates might be expired, but still may be used for checking previously created certificates. By enabling this for all packages (which was at first), will result in people getting a message that they can't resolve (deleting a file is not suggested, but they can't update the certificate either).

So I totally understand your reasoning for including them, but it is not a safe default for the average user.

Idea: what if we create an option (in scan profile) to define if certificates installed by packages should be scan? That could be `ssl-certificate-include-packages` (which would be default set to 'no').

topimiettinen:
I see. Perhaps it should be a bug for a CA to create certificates that extend beyond their own certification and if so, there could be a new test for that.

But the option makes sense. I'll update the PR.

topimiettinen:
Updated

mboelen:
Merged, thanks!

PR #883, opened 2020-03-27
Check if system uses encrypted swap devices

topimiettinen:
Add test CRYP-7931 to check if the system uses any encrypted swap
devices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Using the relevant options, the scan will change base on the intended goal.
 - New test: BOOT-5109 - test presence rEFInd boot loader
 - New test: BOOT-5264 - run systemd-analyze security
 - New test: CRYP-7930 - test for LUKS encryption
+- New test: CRYP-7931 - determine if system uses encrypted swap
 - New test: CRYP-8004 - presence of hardware random number generator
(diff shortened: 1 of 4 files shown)


topimiettinen:
@mboelen Any comments? Is there a problem with the PR?

topimiettinen:
From hardening perspective, unencrypted swap could contain sensitive information, like passwords, encryption keys or traces of activities done in in the system. Swap which is encrypted with a random throw-away key for each boot makes this information unreadable (or more precisely, encrypted but the key no longer exists) for previous boots.

In the forensic case, it's also important to know about this, since it means that rebooting the system will destroy the current contents of the swap. So it may make sense to preserve the contents of the swap first, or not to waste resources to investigate the contents of the swap if the machine was freshly booted up just for the forensic analysis or the disks have been moved to another system.

It's of course possible that no swap will ever be used if there's no pressure from overcommitting of memory.

It would be also possible to use non-random encrypted swap like other encrypted filesystems, the system then needs the encryption key on boot.

Out of scope of this check, but the swap may also be used for resuming the system from disk and then the key can't be random. On new systems the resume device can be separate from swap. The contents of the resume device could be very interesting for forensics.

mboelen:
Looks good to me, so merged!

PR #884, opened 2020-03-28
README.md: add repology badge

topimiettinen:
Show downstream usage for various versions of Lynis.


mboelen:
While the link is useful (using it myself to do a check now and then), I'm wondering about the value to the users. Typically they use the default repo of their distribution.

topimiettinen:
It may show to newcomers that Lynis is already well established software. But I don't mind if you reject the PR.

mboelen:
Merged and moved it a bit to section below regarding software packages.

PR #886, opened 2020-03-29
Fixes for virtual machine dmidecode

bginsbach:
Supports systems install `dmidecode` in _other_ locations by expanding the
_search path_ for `dmidecode` to include more than `/usr/bin`
and `/usr/sbin`. Can not use `${DMIDECODEBINARY}` as this test could
run with `${CHECK_BINARIES} or CORE-1000 being disabled.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -1774,11 +1774,13 @@
         # dmidecode
         # Values: VMware Virtual Platform / VirtualBox
         if [ -z "${SHORT}" ]; then
-            if [ -x /usr/bin/dmidecode ]; then DMIDECODE_BINARY="/usr/bin/dmidecode"
-            elif [ -x /usr/sbin/dmidecode ]; then DMIDECODE_BINARY="/usr/sbin/dmidecode"
-            else
-                DMIDECODE_BINARY=""
-            fi
+            DMIDECODE_BINARY=""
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for pointing this out. Instead of doing a search through the paths, it is better using the builtin `command` functionality of the shell. So won't apply this pull request, but instead did a quick rewrite based on your correct findings.

Commits
https://github.com/CISOfy/lynis/commit/53ad72e791ea378e4974e2db1eb416d4ee537455
https://github.com/CISOfy/lynis/commit/a38e2b535e09ec12e58a9243ef942c9f778acb7a

Let me know if my changes do not work as intended ;-)

mboelen:
Closing pull request (see remarks above)

bginsbach:
Considered using `command` but it was not clear that all versions of shell (e.g. BusyBox) were going to support `command`. Also followed the pattern used in  CORE-1000.

mboelen:
FYI, CORE-1000 has a special function to detect many tools in 1 big sweep. So we tried to make it as quick as possible (and scalable). For that reason, it might be less interesting to just find a single binary.

With your good feedback this area has been improved, even though it was a slightly different solution :)

PR #887, opened 2020-03-29
AUTH-9229 Do not use long options for sort

bginsbach:
Use the standard `sort(1)` short option `-u` rather than `--unique`,
since not all versions support long options.


mboelen:
Good finding. It is unfortunate that long options don't work for all systems (= better readability), so let's fall back on the short one for this. Merged!

PR #889, opened 2020-03-29
PROC-3802 Only check for prelink package on Linux

bginsbach:
The prelink package is Linux specific no need to check for it on
non-Linux systems.


mboelen:
Thanks, that makes sense. Merged!

PR #890, opened 2020-03-31
Add pkg_info

bginsbach:
These changes will remove numerous errors when `PackageIsInstalled` is called on a system using pkgsrc (e.g. NetBSD). It still does not solve the underlying issue with `PackageIsInstalled` doesn't work correctly on BSD systems where the core/base commands/daemons are not tracked as packages but considered part of the (core) OS.


mboelen:
Perfect, it is merged!

PR #893, opened 2020-03-31
Enhance binaries report

topimiettinen:
Report also number of set-uid and set-gid binaries found.

### DIFF
--- a/include/binaries
+++ b/include/binaries
@@ -74,6 +74,10 @@
             fi
         done
 
+        NSUID_BINARIES=0
+        NSGID_BINARIES=0
+        SUID_BINARIES=
+        SGID_BINARIES=
         # Now perform binary detection
         for SCANDIR in ${BIN_PATHS}; do
             SCANDIR=$(echo "${SCANDIR}" | sed 's/!!space!!/ /g')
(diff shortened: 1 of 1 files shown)


Schmuuu:
Hi, 
What do you think about additional information in the "show details" to help the user figure out which binaries have these bits set?
I think about either listing the binaries, which might clutter the log or tell the user how he can find out himself which binaries are "affected".

topimiettinen:
Good idea! I think similar things are logged elsewhere too.

topimiettinen:
Updated with your suggestion.

I've been thinking about detecting non-PIE executables and other flags (execstack). The only decent tool I've found is Debian hardening-check, but it is very slow.

topimiettinen:
Another idea for adding to `tests_hardening` could be to check if on a 64 bit system there are 32 bit libraries or if (on Linux) non-native binary types have been registered in `/proc/sys/fs/binfmt_misc`. The threat would be that the non-native emulations might be buggy.

mboelen:
Thanks, merged the pull request. Made a small edit: call Report for each individual counter (instead of all on one line).

PR #894, opened 2020-04-01
Skip VM detection for show helper

bginsbach:
No need for VM detection when running the `show` helper. This cleans
up errors caused when `$AWKBINARY` is not set as `show` also does
not check for binaries.

### DIFF
--- a/include/parameters
+++ b/include/parameters
@@ -160,6 +160,7 @@
                 RUN_TESTS=0
                 RUN_UPDATE_CHECK=0
                 SKIP_PLUGINS=1
+                SKIP_VM_DETECTION=1
                 SHOW_PROGRAM_DETAILS=0
                 SHOW_TOOL_TIPS=0
                 shift; HELPER_PARAMS="$@"


mboelen:
The detection is required for "lynis show environment" to display details about its host system.

Can you share the error that you get?

bginsbach:
Here is an example.
```
host> ./lynis show categories
./lynis: -v: not found
./lynis: -v: not found
./lynis: -v: not found
./lynis: -v: not found
./lynis: -v: not found
performance
privacy
security
```
Here is an example when running with `sh -x`.
```
+ LogText 'Performing ps scan using partial match and without uid'
+ '[' '!' /tmp/lynis.log = '' -a 0 -eq 1 ']'
+ ps
+ -v pattern=vmware-guestd '$0 ~ pattern {print}'
./lynis: -v: not found
+ FIND=''
```

Again, this is all because`${AWKBINARY}` isn't set since the binary check isn't run.

mboelen:
Thanks, that is helpful. I can't reproduce this yet on my test system. Most likely as it does hit the `IsRunning` function near the end of the virtual machine detection.

Does your system has `pgrep`? I'm wondering why ${PSBINARY} is working/filled in your case, as I expected that one to bail out as well.

bginsbach:
Yes the system used has `pgrep`. Again, when ${AWKBINARY} is empty, as it will be at this point, and the shell tries to exec `-v` as a command.

Here is a full `sh -x` capture.
[lynis.full-debug.txt](https://github.com/CISOfy/lynis/files/4416162/lynis.full-debug.txt)

mboelen:
Looks like `pgrep` (${PGREPBINARY}) was not used, as the variable was empty as well (as expected). Now I see why `ps` (${PSBINARY}) works fine, as it is defined in `include/const` to have a default.

I made a change to the IsRunning function. Can you have a look on your system with the latest code to see if that resolves it?

bginsbach:
No it didn't resolve; just changed the symptom.

```
host> ./lynis show categories
./lynis: \n: not found
./lynis: \n: not found
./lynis: \n: not found
./lynis: \n: not found
./lynis: \n: not found
performance
privacy
security
```

mboelen:
I did some poking and found two more missing entries. I manually enforced it to do the IsRunning check (from IsVirtualMachine) and worked for me. Does it now work for you as well?

Related commit https://github.com/CISOfy/lynis/commit/b5a2d11738cf72691f3b09c48a4c647a4c499277

bginsbach:
Yes that fixes the immediate issue. There could still be other cases lurking where a default binary is required.  I will close this PR.

PR #895, opened 2020-04-01
WIP: Check for non-PIE ELF binaries

topimiettinen:
Check if there are ELF binaries which are not compiled as Position
Independent Executable (PIE).

### DIFF
--- a/include/binaries
+++ b/include/binaries
@@ -74,10 +74,18 @@
             fi
         done
 
+        if [ -x /bin/readelf -o -x /usr/bin/readelf -o -x /usr/local/bin/readelf ]; then
+            CHECK_PIE=1
+        else
+            CHECK_PIE=0
+        fi
+
         NSUID_BINARIES=0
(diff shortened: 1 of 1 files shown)


topimiettinen:
I'm not too happy with this. By using `readelf`, it's faster than `hardening-check` (which also is not widespread), but it still takes more than 10 seconds.

Also, if there are some non-PIE binaries (like GCC stuff and Emacs in my system), what can the system administrator do about it? If the vendor does not provide PIE binaries then it's unlikely that the system administrator can fix it locally by recompiling everything.

mboelen:
Agree, this is not something the average user will do. And if the testing takes this long, then most likely it is not worth the effort considering there are many binaries. Probably better to put this testing into a standalone tool?

topimiettinen:
For Debian, there's already a Lintian check and `hardening-check` as part of `devscripts` package. Perhaps other distros and BSDs have other checks. I'll close this.

PR #898, opened 2020-04-03
AUTH-9268 Add DragonFly

bginsbach:
DragonFly also supports PAM. Rework to use the `--os` option of `Register`
rather than `--preqs-met` as the former can support a list.


mboelen:
Thanks, and also I do like the approach (less lines, still clear to read). Merged!

PR #899, opened 2020-04-03
AUTH-9218 Improvements

bginsbach:
All of the BSD based OSes have `/etc/master.passwd`. Make AUTH-9218 work for all removing
redundant AUTH-9489 for DragonFly. For all BSD OSes `/etc/master.passwd` is only readable
by root so additional condition to test registration.


mboelen:
Perfect change, makes sense. Merged!

PR #900, opened 2020-04-04
Check for IOMMU

topimiettinen:
Add a new test to check if I/O MMU is enabled in the kernel command
line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ Using the relevant options, the scan will change base on the intended goal.
 - New test: FINT-4341 - verify status of dm-verity (Linux)
 - New test: INSE-8314 - test for NIS client
 - New test: INSE-8316 - test for NIS server
+- New test: KRNL-5731 - check if IOMMU is enabled
 - New test: NETW-2706 - check DNSSEC (systemd)
 - New test: NETW-3200 - determine enabled network protocols
 - New test: PHP-2382 - detect listen option in PHP (FPM)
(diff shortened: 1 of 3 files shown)


topimiettinen:
I thought about enhancing this with a recommendation on how to enable the IOMMU. The kernel command line parameter could be simply `intel_iommu=on` for Intel processors but something like  `amd_iommu=[on | fullflush | force_isolation ...]` for AMD.

Also it would be nice to read dmesg, /proc/kmsg, journal or log files to see if any IOMMUs actually were detected and successfully enabled.

Varbin:
Out of curiosity: How is it with ARM or other platforms? What about lower end (or old) processors / chipsets not having one?

For an administrator without knowledge of the hardware capabilities it would not be helpful to see it as "not enabled" if there is no IOMMU.

topimiettinen:
From the Linux kernel sources it seems that many architectures may have an IOMMU including ARM/ARM64. I don't know how it could be enabled there, [kernel-parameters.txt](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/kernel-parameters.txt) does not tell that. Perhaps that file should be given as a pointer. I had overlooked also the fact that intel_iommu= can have other options than just 'on'.

I didn't find a simple way to check if IOMMU is present or enabled. Scraping the logs could show something but the messages are different for just Intel and AMD, no idea about other architectures.

The threat model is that a hardware device could be used to read or write physical memory, bypassing every restriction set by the operating system. The device could be something that is plugged in by the attacker like Firewire (so the attacker needs physical access aka Evil Maid scenario), or it could be an existing device with some processing power, which could be taken over by external attacker (for example GPU or network device). IOMMU limits the area of physical memory which the devices can access to just to those which the OS allows.

mboelen:
I gave this pull request some more thought than others, because at this moment I think we should not include it yet. The primary reason is that it is hard to give the user good advice to make an informed decision. As you also explained, there are multiple options (not just on/off), making it even more complicated.

So for now I will close it, so we can fully focus on the major 3.x release. Later on, we can readdress hardware checks, including this one.

PR #905, opened 2020-04-10
Check for registered non-native binary formats

topimiettinen:
Examine /proc/sys/fs/binfmt_misc (Linux) for additional registered
binary formats. Those are probably emulated and their emulation could
be less tested, more buggy and more vulnerable than native binary
formats, so they should be disabled when not needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ Using the relevant options, the scan will change base on the intended goal.
 - New test: FINT-4316 - presence of AIDE database and size test
 - New test: FINT-4340 - check dm-integrity status (Linux)
 - New test: FINT-4341 - verify status of dm-verity (Linux)
+- New test: HRDN-7231 - check for registered non-native binary formats
 - New test: INSE-8314 - test for NIS client
 - New test: INSE-8316 - test for NIS server
 - New test: NETW-2400 - test hostname for valid characters and length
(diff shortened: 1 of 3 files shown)


topimiettinen:
I think BSDs also have ways to execute at least Linux binaries, but I don't know the details.

Varbin:
@topimiettinen FreeBSD [has an emulation layer](https://www.freebsd.org/doc/handbook/linuxemu-lbc-install.html) for Linux where a kernel module and some libraries must be installed for it. OpenBSD certainly has not this capability (was afaik a feature removed very early), NetBSD [has a layer for FreeBSD, Linux, Solaris and Ultrix](https://www.netbsd.org/docs/compat.html).

Anyway, under `/proc/sys/fs/binfmt_misc/*` I do not only find emulated binaries (`qemu-[architecture]`) but also Java files (`jar`), Python (`python[version]`) and LLVM (`llvm-9-runtime.binfmt`, should not be on servers).

topimiettinen:
Thanks for the links. They don't describe how to detect the emulation, so perhaps someone who knows the layers better can contribute checks for those.

I don't think anything really relies on binfmt_misc for Java or Python integration. At least on my system (where I've disabled mounting the binfmt_misc thing), Java and Python programs run just fine.

topimiettinen:
Unstale

topimiettinen:
ping @mboelen

topimiettinen:
@mboelen Could you pretty please check this PR, it was already closed by the bot because of inactivity. Ditto for #911, #914 and #926. #933 and #936 are still open but who knows how long.

mboelen:
Merged, thanks!

PR #911, opened 2020-04-18
Check mount options of tmpfs and devtmpfs filesystems

topimiettinen:
Sum of sizes of all tmpfs and devtmpfs mounts should be less than the
size of the RAM to prevent out of memory situations.  Without the
size= mount option, default size of each mount is 50% of RAM.


topimiettinen:
BSDs also seem to have tmpfs with similar size= option, so the OS set could be widened later.

topimiettinen:
Also nr_inodes= should be checked.

topimiettinen:
Updated to check for nr_inodes=.

topimiettinen:
@mboelen Is there something to be improved here?

PR #914, opened 2020-04-25
[MACF-6234] Check for Netfilter SECMARK rules

topimiettinen:
SELinux can optionally use Netfilter SECMARK rules to label
packets. Use nft tool or less preferably iptables to check for SECMARK
rules.


topimiettinen:
@mboelen Sadly there are not so good docs for SECMARK besides [SELinux Wiki](https://selinuxproject.org/page/NB_Networking). Current SELinux reference policy will install example NFT and iptables filter files and with the older versions the script to generate them could have been found in the sources for a long time. After the NFT/iptables filters have been installed, TE rules for packet class can be used to further contain any domains. This in practice can be used for application specific network access control (like per-application firewalls), which isn't easily available on Linux otherwise.

PR #920, opened 2020-05-04
Fix KRNL-5730 to properly check /proc/config.gz

jsrc27:
When KRNL-5728 locates the kernel config it does not properly set LINUXCONFIGFILE
if config is found as /proc/config.gz. This causes KRNL-5730 to fail due to missing prereqs,
despite a kernel config existing.


mboelen:
Thanks, good catch! Merged!

PR #926, opened 2020-05-15
New test: KRNL-6001 - check SysV shared memory configuration

topimiettinen:
Check that Linux sysctl kernel.shmall is less than RAM size to prevent
out of memory situations.


topimiettinen:
@mboelen Is there anything to be improved here?

PR #936, opened 2020-05-25
[NETW-3200] Check lots of uncommon protocols

topimiettinen:
Check lots of uncommon protocols. Adjust reporting to not print a list
of every possible protocol since the list is now longer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,7 @@ Using the relevant options, the scan will change base on the intended goal.
 - NAME-4404 - improved screen and log output
 - NAME-4408 - corrected Report function call
 - NETW-3032 - small rewrite of test and extended with addrwatch
+- NETW-3200 - check lots of uncommon protocols
 - PHP-2372 - don't look in the cli configuration files
 - PKGS-7388 - only perform check for Debian/Ubuntu/Mint
 - PKGS-7410 - use multiple package managers when available
--- a/include/tests_networking
+++ b/include/tests_networking
@@ -730,11 +730,12 @@
     if [ ${SKIPTEST} -eq 0 ]; then
         TESTED=0
         FOUND_UNCOMMON_PROTOCOL_ENABLED=0
+        FOUND_UNCOMMON_PROTOCOLS=""
         case ${OS} in
             Linux)
                 TESTED=1
                 LogText "Test: checking the status of some network protocols that typically are not used"
-                UNCOMMON_PROTOCOLS="dccp sctp rds tipc"
+                UNCOMMON_PROTOCOLS="af_packet_diag appletalk atm batman_adv can ceph dccp dccp_diag decnet diag ife l2tp_ip l2tp_ip6 l2tp_netlink l2tp_ppp lapb llc llc2 mpls_router netlink_diag netrom openvswitch phonet pktgen pn_pep pppoe ppox pptp raw_diag rose rxrpc scsi_transport_iscsi sctp sctp_diag smc smc_diag rds tipc unix_diag vsock vsock_diag"
                 for P in ${UNCOMMON_PROTOCOLS}; do
                     LogText "Test: now checking module '${P}'"
                     if ! SkipAtomicTest "${TEST_NO}:${P}"; then
@@ -759,8 +760,7 @@
                         fi
 
                         if [ ${UNCOMMON_PROTOCOL_DISABLED} -eq 0 ]; then
-                            ReportSuggestion "${TEST_NO}" "Determine if protocol '${P}' is really needed on this system"
-                            Report "uncommon_network_protocol_enabled=${P}"
+                            FOUND_UNCOMMON_PROTOCOLS="${FOUND_UNCOMMON_PROTOCOLS} ${P}"
                             FOUND_UNCOMMON_PROTOCOL_ENABLED=1
                         fi
                     fi
@@ -774,12 +774,14 @@
         esac
         if [ ${TESTED} -eq 1 ]; then
             if [ ${FOUND_UNCOMMON_PROTOCOL_ENABLED} -eq 1 ]; then
-                Display --indent 2 --text "- Uncommon network protocols" --result "${FOUND}" --color YELLOW
+                Display --indent 2 --text "- Uncommon network protocols" --result "${STATUS_FOUND}" --color YELLOW
+                ReportSuggestion "${TEST_NO}" "Determine if all network protocols are really needed on this system"
+                Report "uncommon_network_protocol_enabled=${FOUND_UNCOMMON_PROTOCOLS}"
             else
                 Display --indent 2 --text "- Uncommon network protocols" --result "${STATUS_NOT_FOUND}" --color GREEN
             fi
         fi
-        unset DATA F FOUND TESTED UNCOMMON_PROTOCOLS
+        unset DATA F FOUND FOUND_UNCOMMON_PROTOCOLS TESTED UNCOMMON_PROTOCOLS
     fi
 #
 #################################################################################


Varbin:
How do I disable certain protocols? If I e.g. use ceph for storage I do not want to warned if I use it, but maybe still if other protocols are enabled.

_openvswitch_ is quite common for advanced switching functionality, and _pppoe_ is commonly used for authenticating to a DSL modem.
All of them should not be enabled on a regular server, but might be common enough to individually disable them.

When checking for file permissions (/etc/passwd and so on), the filenames are read from the profile instead of being coded in bash.

topimiettinen:
I don't think there is an override for every test but it could be a good idea to reduce noise.

The general suggestion here is to look at all modules and check if there are unneeded modules which could be disabled. The list of modules may change every time a new kernel version is installed. This test just looks at available modules, if the protocols are compiled in or new unknown modules are available, there's no warning. Perhaps instead of checking for specific protocol (and file system) modules and giving a list of suggestions for removal (which may never be accurate anyway), there could be simple statistics like `questionable_modules = (loaded_modules + blocked_modules) / (total_modules) * %`. This would tell if there are a lot of unused modules (not just file systems and protocols) which are not blocked. For example, on my system the number is 9.7% which seems awful.

Varbin:
> I don't think there is an override for every test but it could be a good idea to reduce noise.

Do you refer to your module or any other test? Whole tests and even subtests can be skipped in the profile files (by using `skip-test=TEST-ID:subtest`; Specifying `subtest` is optional). This should work for all tests available.
I hope this can be easily extended to this module, as I do not know how this is implemented (the test checking the OpenSSH config should have this, so you might want to look there).

> The general suggestion here is to look at all modules and check if there are unneeded modules which could be disabled.

Do you mean checking all unused kernel modules on a system?

I don't know. Many of the unused modules are hardware specific, and are only used if the respective hardware is present in the system -- I am not sure if (and how) unloaded modules are a risk.
Only root can manually load modules, and modules automatically loaded after an easy action by an attacker (network, usb, firewire, filesystem) are currently checked for.

It might even have unforeseen consequences to disable / blacklist unused modules. I say this, as I ran a custom kernel, until a physical keyboard had to be used, but didn't work as both the HID and the USB modules were removed (you might want to read Microsoft's essay [Sticking with Well-Known and Proven Solutions](https://docs.microsoft.com/en-us/archive/blogs/fdcc/sticking-with-well-known-and-proven-solutions)).

>  Perhaps instead of checking for specific protocol (and file system) modules and giving a list of suggestions for removal (which may never be accurate anyway), there could be simple statistics like questionable_modules = (loaded_modules + blocked_modules) / (total_modules) * %.

I do not think showing the administrators / users such a number is helpful for them. 

_Don't get me wrong_: You are essentially right.
There are many unused modules that are almost never used (like appletalk in your PR), and some certainly are unmaintained but loaded in some cases.
I just think extending this to all modules goes to far and can prove to be unrealistic or unhelpful.

topimiettinen:
> Do you refer to your module or any other test? Whole tests and even subtests can be skipped in the profile files (by using skip-test=TEST-ID:subtest; Specifying subtest is optional). This should work for all tests available.
> I hope this can be easily extended to this module, as I do not know how this is implemented (the test checking the OpenSSH config should have this, so you might want to look there).

Yes, something like that, skip checks for modules X & Y. Perhaps also the opposite could make sense: only run checks for modules A, B & C. 

> 
>     The general suggestion here is to look at all modules and check if there are unneeded modules which could be disabled.
> 
> Do you mean checking all unused kernel modules on a system?
> 
> I don't know. Many of the unused modules are hardware specific, and are only used if the respective hardware is present in the system -- I am not sure if (and how) unloaded modules are a risk.
> Only root can manually load modules, and modules automatically loaded after an easy action by an attacker (network, usb, firewire, filesystem) are currently checked for.
> 
> It might even have unforeseen consequences to disable / blacklist unused modules. I say this, as I ran a custom kernel, until a physical keyboard had to be used, but didn't work as both the HID and the USB modules were removed (you might want to read Microsoft's essay Sticking with Well-Known and Proven Solutions).

Disabling modules, or more generally, doing any kind of hardening can make services fail or make the system unbootable. Reducing the attack surface may also accidentally take away something useful. I think all recommendations by a humble tool like Lynis (or just trying some hardening stuff found in the net or reading books) have to be understood to a degree (an opportunity for learning something) and then tested well if they make sense in a system in question. Classical mistake is to make mistakes working with a remote system and then having to physically move to the location to fix the problem (or worse).

The advantage of Linux is that it's relatively easy to try something and if it doesn't work, try to find out why it fails and then fix it, or just revert. An expert (maybe just somebody who has tried and failed enough) can make changes like this with confidence. Especially with Windows everything is so opaque that if there's a major problem, it's often easiest to reinstall, so no wonder MS recommends sticking to simple solutions.

>     Perhaps instead of checking for specific protocol (and file system) modules and giving a list of suggestions for removal (which may never be accurate anyway), there could be simple statistics like questionable_modules = (loaded_modules + blocked_modules) / (total_modules) * %.
> 
> I do not think showing the administrators / users such a number is helpful for them.

There's also an error, the ratio isn't about "questionable modules".

> Don't get me wrong: You are basically right.
> There are many unused modules that are almost never used (like appletalk in your PR), and some certainly are unmaintained but loaded in some cases.
> I just think extending this to all modules goes to far and can prove to be unrealistic or unhelpful.
> 
> Therefore I suggest making it possible to skip each individual module and restrict testing them to network protocols only.

I agree with you somewhat as checking for any modules and then blacklisting them is a bit of game of whack-a-mole. I also don't know if DCCP is more dangerous than Batman or when either could be useful. But in a server setup, instead of allowing 1700, 1699 or 1698 modules to be loaded, more useful advice would be that the set of modules should be minimized and fixed by disallowing module loading, after initial set of known, used and necessary modules have been loaded. This is of course also harder to implement by the administrator than just blacklisting some modules which probably are not used. This also depends on what's the goal of Lynis, is it to provide suggestions which are easy to implement and sort of lowest common denominator but perhaps not so effective? Can we assume that the administrators reading the report from Lynis knows roughly what's PPoE and if the systems use it or not before blacklisting the module, or they have fingers burnt enough to know to get informed before making the decision? But anyway, that should be a separate PR.

mboelen:
This one was on 'hold' as I wanted to give it a good thought. Actually, I think you are both right. Yes, one should not just turn off things because a tool is saying so. At the other hand, it is a tool like Lynis that can help with spotting room for improvement.

What if we introduce a "complexity level" of 1 to 5. By default, it can be set to something like 3, which means most current tests will be active. However, if you turn it up to 4 or 5, then some tests will be activated or they will be more strict. A good example of tests would be this one. A test level like 1 would be for basic measures like a required reboot or dealing with warnings. Level 2 would ease up on some of the existing tests, like SSH settings to include only a few basic ones. Level 3 would cover most of the current ones, where 4 and 5 go all the way.

What are your thoughts @Varbin @topimiettinen ?

topimiettinen:
I like the complexity idea. For example, locking down all modules after loading is difficult (I'm finding problems every now and then with this and kernel is rather asynchronous with modules, so something which works once may fail on next boot) but disabling a limited list of obscure modules is much easier.

Some hardening tweaks may require the admin to set a numeric limit but figuring out a good value depends on understanding some other parts of the system, so those could be on more advanced level.

The ultimate level could target developers who may be powerful enough to recompile packages, even kernel. For example, previously dismissed PIE (position independent executables) check could fall to this criteria. SELinux can be also tricky, in the worst case the distro does not even support it fully.

Varbin:
@mboelen Regarding the complexity levels: I think this is a great idea.
When currently running Lynis, one gets a lot of suggestions which are not weighted much in terms of severity (internally there are hardening points and the two possible levels now) and complexity to fix. Especially fixing high complexity / low severity problems takes a tremendous amount of time, and correctly identifying them takes some experience.

Having complexity labels, would allow easily identifying "quick wins" that can be fixed in close to no time. 

Maybe [we can create a discussion for this](https://github.com/CISOfy/lynis/discussions/new)?

PR #951, opened 2020-06-20
Update language files (de, de-AT, en)

al-lac:
Updated the german language file with new strings and created a de-AT link to it. 

Reordered the english language file alphabetically so it is easier to find strings and to compare to other language files.


al-lac:
You are right. The VS Code Sort Plugin does not take this into consideration. 

I changed it.

mboelen:
Merged, thanks!

PR #957, opened 2020-06-20
rsh host file permissions

Varbin:
The old rsh (remote shell) authorizes users and hosts in the files `/etc/hosts.equiv`, `.rlogin`and `.rhosts`. If attackers can write to those files, he can logon as a different to the system. While the rsh daemon usually checks for non-root owners or write permission, this may not be the case on any system.

Those files are also used by SSH if host based authentication is enabled. In addition to the files used by rsh, `.shosts` and `shosts.equiv` are used.

While the use of rsh is highly discouraged, it may still be in running on legacy systems. As those files are also used by OpenSSH, they are even relevant today.

This PR adds those files to the list of checked files.


mboelen:
Thanks, merged!

PR #958, opened 2020-06-21
fix stderr output from cryptsetup status

Steve8291:
Redirected stderr to /dev/null to silence output of `cryptsetup status /swap.img`
This was causing error output from my cron script.
Otherwise, if the swap file is not encrypted then the following error will be printed:
`Device swap.img not found`


Steve8291:
I have tested with Lynis 3.0.0 and this works without any problems.

mboelen:
Thanks, that makes sense. Merged!

PR #969, opened 2020-06-26
[AUTH-9229] resolving syntax error on MacOS Catalina

### LINKED ISSUE #968: syntax error in tests_authentication on MacOS Catalina
On MacOS, lynis reports a syntax error in `include/tests_authentication`, which causes many of the tests defined there to be skipped.

**Version**
 - MacOS Catalina, 10.15.5
 - Lynis version: 3.0.1

**Expected behavior**
No syntax error, and all of the tests in include/tests_authentication should be run.

**Output**
```
[+] Users, Groups and Authentication
  - Administrator accounts                                    [ OK ]
  - Unique UIDs                                               [ OK ]
  - Unique group IDs                                          [ OK ]
  - Unique group names                                        [ OK ]
include/tests_authentication: line 307: syntax error near unexpected token `;;'
include/tests_authentication: line 307: `                    ;;'

[+] Custom tests
  - Running custom tests...                                   [ NONE ]
```

**Additional context**
The syntax error is in the case statement used in test AUTH-9229, and is almost certainly happening because MacOS Catalina is using an old version of bash in lieu of an actual bourne shell, i.e. `/bin/sh --version` reports that it's actually GNU bash 3.2.51(1)

Refactoring the case statement into its own function appears to resolve this problem; I'll submit a pull request for that once I've submitted this issue.

ww2:
Fixes #968 by moving the existing case statement in AUTH-9229 into its own function.

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -295,50 +295,56 @@
             PREQS_MET="YES"
         ;;
     esac
+
+    function ParsePasswordEntry() {
+        METHOD=$1
+        case ${METHOD} in
+            1:\* | 1:x | 0: | *:!* | *LOCK*)
+                # disabled | shadowed | no password | locked account (can be literal *LOCK* or something like LOCKED)
+                ;;
+            *:\$5\$*| *:\$6\$*)
+                # sha256crypt | sha512crypt: check number of rounds, should be >5000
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks! I was troubleshooting it and could not reproduce and resolve it. Definitely makes sense now.

Issue #964, opened 2020-06-25
Unknown OS found when running on Pop!_OS


bobzurad:
**Describe the bug**
When running lynis on Pop!_OS, I get the following message included in the output:

  Exception found!

  Function/test:  [OS Detection]
  Message:        Unknown OS found in /etc/os-release - Please create issue on GitHub project page: https://github.com/CISOfy/lynis

Here is my output of cat /etc/os-release:
NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.04 LTS"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os

**Version**
 - Distribution: Pop!_OS 20.04
 - Lynis version: 3.0.1

**Expected behavior**
Lynis should recognize Pop!_OS as a valid OS.

chrislynch8:
Found this tool this morning and it does not find POP!_OS which I run. Quick change made to include Pop!_OS in the os detection script

mboelen:
With change of @chrislynch8 it will be detected in next release. Thanks both for the help.

PR #970, opened 2020-06-27
Fix for Issues #964 - Pop!_OS added to osdetection

chrislynch8:
Fix for https://github.com/CISOfy/lynis/issues/964 - POP OS added to include/osdetection


mboelen:
Thanks @chrislynch8 - It has been merged!

PR #972, opened 2020-07-02
[FILE-6430] Don't grep nonexistant modprobe.d files

kepi:
I think we don't want to grep files in modprobe.d when dir is empty. Uses same approach as in USB-1000.


mboelen:
Thanks, merged!

PR #973, opened 2020-07-02
Support absolute paths in nginx includes and fix ls warning on empty directories

kepi:
Includes can be absolute paths too. This is quick fix counting on fact that absolute paths have slash at start.

Second commit is similar to my previous PR, IMHO Lynis should not output ls errors on empty included directories.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -2305,7 +2305,8 @@
                     # Check for additional config files included as follows
                     # "include sites-enabled/*.conf"
                     elif [ $(echo ${VALUE} | grep -F -c "*.conf") -gt 0 ]; then
-                        for FOUND_CONF in $(ls ${CONFIG_FILE%nginx.conf}${VALUE%;*}); do
+                        if [ "$(echo ${VALUE} | ${CUTBINARY} -c1)" != "/" ]; then VALUE=${CONFIG_FILE%nginx.conf}; fi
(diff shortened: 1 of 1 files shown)


include/functions
```diff
@@ -2305,7 +2305,8 @@
                     # Check for additional config files included as follows
                     # "include sites-enabled/*.conf"
                     elif [ $(echo ${VALUE} | grep -F -c "*.conf") -gt 0 ]; then
-                        for FOUND_CONF in $(ls ${CONFIG_FILE%nginx.conf}${VALUE%;*}); do
+                        if [ ${VALUE:0:1} != '/' ]; then VALUE=${CONFIG_FILE%nginx.conf}; fi
```
mboelen, line 2308: This looks like it is not POSIX compatible. Can you rewrite it using `if [ "$(echo ${VALUE} | ${CUTBINARY} -c1)" != "/" ]` and test it that works correctly on your system?

kepi: Totally, I was afraid it might be the case but then lazy tested it in dash and thought it would be ok. I have to check for some option to check for non-posix constructions automatically.

Anyway rewritten and tested on 100+ systems, everything looks ok (in our environment).

kepi:
I'm not sure how to confirm that requested changed was done long time ago. This should wait only for review IMHO.

mboelen:
Thanks, merged!

PR #974, opened 2020-07-02
Command line option for slow test threshold

kepi:
I believe it should be OK to run long tests if we count with it.

Example:

    lynis audit system --slow-warning 300

Will warn when test takes longer than 300 seconds, instead of default 10.


mboelen:
Great suggestion and change, so merged!

PR #980, opened 2020-07-09
Fix timesyncd detection on systems without dbus.

### LINKED ISSUE #953: systemd-timesync test shows a warning if dbus is not installed
**Describe the bug**
It is possible to use systemd-timesyncd without dbus, and Debian allows its users to remove it. Without dbus, timedatectl does not work but prints and error.

**Version**
 - Raspberry Pi OS Buster (formerly raspbian)
 - 3.0.1, bd29a3e4e790d9825521ae9e9a04ffb3c01721d0

**Expected behavior**
I'm not sure. Show something like "systemd-timesyncd running, but it is not possible to theck the status"?

At least the presence should be detected (systemd-timesyncd is installed) and if the service runs. Lynis should note that there is a problem querying it. 

**Output**

```
[+] Time and Synchronization
Failed to create bus connection: Connection refused
  - Checking for a running NTP daemon or client               [ WARNING ]
```

```
  * Use NTP daemon or NTP client to prevent time issues. [TIME-3104] 
      https://cisofy.com/lynis/controls/TIME-3104/
```

Varbin:
Detect timesyncd by looking in the process list instead of timedatectl. The check if time was successfully synchronized is in the new test `TIME-3185`.

Closes #953.

### DIFF
--- a/db/tests.db
+++ b/db/tests.db
@@ -419,6 +419,7 @@ TIME-3170:test:security:time::Check configuration files:
 TIME-3180:test:security:time::Report if ntpctl cannot communicate with OpenNTPD:
 TIME-3181:test:security:time::Check status of OpenNTPD time synchronisation
 TIME-3182:test:security:time::Check OpenNTPD has working peers
+TIME-3185:test:security:time::Check systemd-timesyncd synchronized time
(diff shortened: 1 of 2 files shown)


mboelen:
Thanks, merged!

PR #824, opened 2020-01-08
Add detection of OpenNTPD

Varbin:
This fixes #518.

Summary of changes:
- Enhancements:
  - Search for the `ntpctl`binary.
  - Check if OpenNTPD is installed by checking if `ntpctl` is installed.
  - Confirm the use of OpenNTPD by confirming `ntpctl` is working or an process with the commandline `ntpd: ntp engine` exists.
  - Add typical OpenNTPD configuration files to TIME-3170.
- New tests:
  - TIME-3180: Report if ntpctl cannot communicate with OpenNTPD
  - TIME-3181: Check status of OpenNTPD time synchronisation
  - TIME-3182: Check OpenNTPD has working peers

This has been tested on FreeBSD

PR #981, opened 2020-07-09
[TIME-3180, TIME-3181, TIME-3182] Fix OpenNTPD tests

Varbin:
In #824 I made a few mistakes:

 - All tests:
   - To test if the test should run, I checked against `NTPD_RUNNING` instead of `NTP_DAEMON_RUNNING`.
   - Additionally, I used `==` instead of  `=` when testing if the daemon is openntpd. While this works on some systems, this is non standard and does not work at all on BSDs (where OpenNTPD is usually installed).
 - TIME-3182:
   - The regex for the peers was wrong (never matched anything) and instead of checking for an empty string (`-z`) it was checked for non empty (`-n`).

Thus the previous implementations were completely broken (TIME-3180, TIME-3181, TIME-3182). With this pull request I fix my own errors.

### DIFF
--- a/include/tests_time
+++ b/include/tests_time
@@ -100,7 +100,7 @@
                 LogText "result: running openntpd not found, but ntpctl is instaalled"
             fi
 
-            if [ "${NTP_DAEMON}" == "openntpd" ]; then
+            if [ "${NTP_DAEMON}" = "openntpd" ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
Another set of good changes, thanks and merged!

Issue #826, opened 2020-01-12
[TIME-3104] Wrong directory listing is used


patti2:
**Describe the bug**
Upon trying to find NTP cronjobs in /etc/cron.hourly, the directory listing of the pwd is used instead of the correct directory listing.

**Version**
 - Distribution: Debian buster
 - Lynis version: 2.7.5

**Expected behavior**
check for ntpdate or rdate in /etc/cron.hourly and not in pwd

**Output**
(each x represents a different file that is not present in /etc/cron.hourly, but rather in the pwd)

> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
> 2020-01-12 02:29:38 Test: checking for ntpdate or rdate in /etc/cron.hourly/x
[...]

PR #993, opened 2020-07-21
[TIME-3104] Find more time synchronization commands

Varbin:
This pull request detects the use of `sntp` and `ntpdig`.

As I worked at the same area of code discussed in #826, I fixed it as well.


mboelen:
Thanks, merged!

Issue #892, opened 2020-03-31
No new release in 4 months


duck-rh:
**Describe the bug**
Lynis LYNIS test suggest to upgrade because the installed software is more than 4 months old but there are no new releases.

**Version**
 - Distribution Debian
 - Lynis version 2.7.5

**Expected behavior**
The behavior is right but the test is based on a release frequency expectation that is probably not realistic.

mboelen:
Yes it is a known issue and will be resolved with upcoming release. As this is a major release, it took longer than the normal release (usually one new version every month). So closing this issue, as we are getting really close to 3.0.0 ;-)

duck-rh:
@mboelen you closed this ticket before solving it. The fact it is well known does not help at all. More than two months have passed and I see little activity on the git repo, so I guess that was premature.

So I would like this ticket to be reopened, not to put pressure on the release but to discuss how we can make this check better.

mboelen:
@duck-rh The fix has already been applied, so that is why it is closed. We are preparing the new release in the background and the reason that changes are frozen until the release is available.

So the release is coming and the check is already improved.

sazzle2611:
Same issue again, infact has been issue for long time but have been ignoring, the problem is we want to use Lynis to watch for possible security issues but instead I just end up deleting 100s of emails without even looking at them cause all most of them say is that the release is more than 4 months old. This means that I will miss any that are actually important.

We are on version 3.0.8

PR #995, opened 2020-07-24
Do not suggest upgrading if no update is available

rchekaluk:
Lynis 3.0.0 reports the following:
```
  * This release is more than 4 months old. Consider upgrading [LYNIS] 
      https://cisofy.com/lynis/controls/LYNIS/
```

yet lynis.log indicates that no update is available:
```
2020-07-24 10:09:21 Test: Checking for program update...
2020-07-24 10:09:21 Current installed version  : 300
2020-07-24 10:09:21 Latest stable version      : 300
2020-07-24 10:09:21 No Lynis update available.
2020-07-24 10:09:21 Suggestion: This release is more than 4 months old. Consider upgrading [test:LYNIS] [details:-] [solution:-]
```

This pull request omits the suggestion if no update is available.

Related to https://github.com/CISOfy/lynis/issues/892

### DIFF
--- a/lynis
+++ b/lynis
@@ -777,7 +777,7 @@ ${NORMAL}
     RELEASE_PLUS_TIMEDIFF=$((PROGRAM_RELEASE_TIMESTAMP + TIME_DIFFERENCE_CHECK))
     if [ ${NOW} -gt ${RELEASE_PLUS_TIMEDIFF} ]; then
         # Show if release is old, only if we didn't show it with normal update check
-        if [ ${UPDATE_AVAILABLE} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for your pull request, much appreciated!

There are two different messages when it comes to checking:
- Update check performed and user is not running latest version --> suggest upgrading
- Update check NOT performed (e.g. disabled, no connection, otherwise) and user is running a version older than 4 months --> suggest upgrading

So your patch would actually break that last option. Instead, I just changed the text to suggest that one should check manually if there is a newer version available.

The reason you are seeing the message is that 3.0.0 is having an incorrect timestamp. That will be resolved with 3.0.1 release.

Closing this pull request without merging, due to the reason listed above.

PR #1008, opened 2020-08-10
Alpine Improvements

kolenichsj:
Added Alpine Linux EOL dates for current versions with support and most recent version out of support.
Improve detecting kernel version on disk. Alpine uses /boot/vmlinuz-lts which was being detected as version 'lts' which did not match the current version of the kernel in memory. This caused the code to incorrectly identify a reboot as needed.


mboelen:
Thanks, looking good and also thanks for your patience. Merged!

Issue #474, opened 2017-10-12
Gather info on expired passwords and locked accounts


sigio:
It would be nice if lynis would gather (and report in the portal/reports) information about user-accounts:

- Expired or soon to expire passwords (passwd -e)
- Locked accounts (passwd -S )
- Expired accounts (chage -E / chage -l )

justinamcafee:
I'll take a swing at this. Where do I need to store the function? (I'm not great at this, but it looks easy enough to code and produce some formatted output. )

mboelen:
Hi @justinamcafee - You can add it as a separate test (e.g. AUTH-9350), if you are still up for it.

chr0mag:
Much of this is already covered:
- AUTH-9288 checks for expired passwords (but not soon-to-expire passwords)
- AUTH-9286 checks if min/max password aging is enabled system wide
- AUTH-9282 checks for accounts w/out a set password expiry date

Since AUTH-9282/9283 already inspect `passwd --status --all` it would be simple to check for locked 
accounts here as well by adding:
```
FIND3=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="L") print $1 }')
```
...and then adding a new test case (eg. AUTH-9284) to output the locked accounts. 

@mboelen Should system accounts be ignored when listing locked accounts? Most (all?) accounts with 0<UID<1000 and UID<65534 (_nobody_ user) are locked so reporting this may not be helpful.

jotasprout:
I see this was reopened 9 days ago -- what needs to be done?

mboelen:
Sorry for lack of response on this one. Finishing the major 3.x release had priority, which is now completed.

@chr0mag yes, might make sense to ignore system accounts
@jotasprout feel free to pick it up with the suggestion made by @chr0mag

danielorihuela:
Hi guys, I made a PR adding a test for the locked accounts. Is my first PR to the project, any feedback is appreciated.

Thanks.

PR #1009, opened 2020-08-10
[AUTH-9284] Feature: gather locked accounts info

danielorihuela:
Hi guys, I am very excited since this is my very first contribution into this project.

I basically added a test to gather information about locked accounts following the guidelines of @chr0mag as mentiones in issue #474. Furthermore, I removed the system locked accounts as @chr0mag and @mboelen consensued in the same issue.

Any feedback is appreciated.

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -859,23 +859,27 @@
                     PREQS_MET="YES"
                     FIND_P=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="P" && $5=="99999") print $1 }')
                     FIND2=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="NP") print $1 }')
+                    FIND3=$(passwd -a -S 2> /dev/null | ${AWKBINARY} '{ if ($2=="L") print $1 }' | sort | uniq)
                     ;;
                 *)
                     PREQS_MET="YES"
                     FIND_P=$(passwd --all --status 2> /dev/null | ${AWKBINARY} '{ if ($2=="P" && $5=="99999") print $1 }')
                     FIND2=$(passwd --all --status 2> /dev/null | ${AWKBINARY} '{ if ($2=="NP") print $1 }')
+                    FIND3=$(passwd --all --status 2> /dev/null | ${AWKBINARY} '{ if ($2=="L") print $1 }' | sort | uniq)
                     ;;
             esac
         elif [ "${OS_REDHAT_OR_CLONE}" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


danielorihuela:
Thanks for the explanation @Varbin. You are right. I fixed the implementation with your proposal. If you can take a look at it and give me more feedback it would be great! Thanks for your help.

include/tests_authentication
```diff
@@ -859,23 +859,27 @@
...
                     FIND_P=$(passwd --all --status 2> /dev/null | ${AWKBINARY} '{ if ($2=="P" && $5=="99999") print $1 }')
                     FIND2=$(passwd --all --status 2> /dev/null | ${AWKBINARY} '{ if ($2=="NP") print $1 }')
+                    FIND3=$(passwd --all --status 2> /dev/null | ${AWKBINARY} '{ if ($2=="L") print $1 }' | sort | uniq)
                     ;;
             esac
         elif [ "${OS_REDHAT_OR_CLONE}" -eq 1 ]; then
             PREQS_MET="YES"
             FIND_P=$(for I in $(${AWKBINARY} -F: '{print $1}' "${ROOTDIR}etc/passwd") ; do passwd -S "$I" | ${AWKBINARY} '{ if ($2=="PS" && $5=="99999") print $1 }' ; done)
             FIND2=$(for I in $(${AWKBINARY} -F: '{print $1}' "${ROOTDIR}etc/passwd") ; do passwd -S "$I" | ${AWKBINARY} '{ if ($2=="NP") print $1 }' ; done)
+            FIND3=$(for I in $(${AWKBINARY} -F: '{print $1}' "${ROOTDIR}etc/passwd") ; do passwd -S "$I" | ${AWKBINARY} '{ if ($2=="L") print $1 }' | sort | uniq ; done)
```
Varbin, line 875: My CentOS 8 passwd prints *LK* instead of *L*. Maybe both values should be checked here?

Varbin (changes requested):

include/tests_authentication
danielorihuela: Perfect, added. It is starting to worry me how many of this differences might be between different distributions.

danielorihuela: Thanks for you help, do you know if there exists more differences like that. If the answer is yes, maybe we should try and get the least amount of possibilities that cover the most of the linux distributions.

mboelen:
Hi @danielorihuela - Congrats on your first pull request and thanks (also for your patience). I have reviewed it and it has been merged! A few minor changes have been applied, such as showing a suggestion instead of a warning.

With the change being implemented, it will be available in the next release.

PR #1013, opened 2020-08-23
[TIME-3185]: Change timesyncd synchronization file

### LINKED ISSUE #1012: TIME-3185 incorrectly fails on Ubuntu 18.04
**Describe the bug**
On Ubuntu 18.04, which has systemd version 237, the `TIME-3185` test fails while it shouldn't.

Upon inspecting the code I see that it check the `/run/systemd/timesync/synchronized` file for the last synchronization moment:
https://github.com/CISOfy/lynis/blob/4206177081fd05d589548b264bba9fa2eeac3d9f/include/tests_time#L577

However, in systemd version 237, which is the one on Ubuntu bionic, this file does not exist [(see this blog)](https://www.sigmdel.ca/michel/ha/domoticz/domoticz_sync_problem_en.html#domoticz_unit).

A potential fix is to use `/var/lib/systemd/timesync/clock`, which is another file which is modified every time the clock is synchronized [(see the manual)](https://www.freedesktop.org/software/systemd/man/systemd-timesyncd.service.html)

**Version**
 - Distribution: Ubuntu 18.04.5
 - Lynis version: 3.0.1

**Expected behavior**
The `TIME-3185` test should pass

**Output**
```
2020-08-23 20:23:52 Performing test ID TIME-3185 (Check systemd-timesyncd synchronized time)
2020-08-23 20:23:52 Warning: systemd-timesyncd never successfully synchronized time [test:TIME-3185] [details:-] [solution:-]
```

**Additional context**
Below the manual for systemd-timesyncd version 237 in which it is clear that `/run/systemd/timesync/synchronized` is not yet used but `/var/lib/systemd/timesync/clock` **is**:
```
SYSTEMD-TIMESYNCD.SERVICE(8)                                                             systemd-timesyncd.service                                                             SYSTEMD-TIMESYNCD.SERVICE(8)

NAME
       systemd-timesyncd.service, systemd-timesyncd - Network Time Synchronization

SYNOPSIS
       systemd-timesyncd.service

       /lib/systemd/systemd-timesyncd

DESCRIPTION
       systemd-timesyncd is a system service that may be used to synchronize the local system clock with a remote Network Time Protocol server. It also saves the local time to disk every time the clock
       has been synchronized and uses this to possibly advance the system realtime clock on subsequent reboots to ensure it monotonically advances even if the system lacks a battery-buffered RTC chip.

       The systemd-timesyncd service specifically implements only SNTP. This minimalistic service will set the system clock for large offsets or slowly adjust it for smaller deltas. More complex use
       cases are not covered by systemd-timesyncd.

       The NTP servers contacted are determined from the global settings in timesyncd.conf(5), the per-link static settings in .network files, and the per-link dynamic settings received over DHCP. See
       systemd.network(5) for more details.

       timedatectl(1)'s set-ntp command may be used to enable and start, or disable and stop this service.

FILES
       /var/lib/systemd/timesync/clock
           This file contains the timestamp of the last successful synchronization.

SEE ALSO
       systemd(1), timesyncd.conf(5), systemd.network(5), systemd-networkd.service(8), timedatectl(1), localtime(5), hwclock(8)

systemd 237                                                                                                                                                                    SYSTEMD-TIMESYNCD.SERVICE(8)
```

Jimver:
This PR fixes #1012. As Ubuntu 18.04, which has systemd version 237 does not make a file at `/run/systemd/timesyncd/synchronized` I changed the `SYNCHRONIZED_FILE` to `/var/lib/systemd/timesync/clock`, which **is** created by systemd version 237.

This shouldn't cause any issues for later systemd versions as `/var/lib/systemd/timesync/clock` is still used besides `/run/systemd/timesyncd/synchronized` [(see the manual of the latest systemd version)](https://www.freedesktop.org/software/systemd/man/systemd-timesyncd.service.html).

**Update**: I have now modified the PR such that it still uses `/run/systemd/timesyncd/synchronized` first and uses `/var/lib/systemd/timesync/clock` and `/var/lib/private/systemd/timesync/clock` as fallback options in case `/run/systemd/timesyncd/synchronized` doesn't exist.

### DIFF
--- a/include/tests_time
+++ b/include/tests_time
@@ -574,7 +574,7 @@
 
 
     Register --test-no TIME-3185 --preqs-met "${PREQS_MET}" --weight L --network NO --category "security" --description "Check systemd-timesyncd synchronized time"
-    SYNCHRONIZED_FILE="/run/systemd/timesync/synchronized"
+    SYNCHRONIZED_FILE="/var/lib/systemd/timesync/clock"
     if [ ${SKIPTEST} -eq 0 ]; then
         if [ -e "${SYNCHRONIZED_FILE}" ]; then
            FIND=$(( $(date +%s) - $(${STATBINARY} -L --format %Y "${SYNCHRONIZED_FILE}") ))
--- a/include/tests_time
+++ b/include/tests_time
@@ -574,8 +574,17 @@
 
 
     Register --test-no TIME-3185 --preqs-met "${PREQS_MET}" --weight L --network NO --category "security" --description "Check systemd-timesyncd synchronized time"
-    SYNCHRONIZED_FILE="/var/lib/systemd/timesync/clock"
+    SYNCHRONIZED_FILE="/run/systemd/timesync/synchronized"
+       
     if [ ${SKIPTEST} -eq 0 ]; then
+        # On earlier systemd versions (237), '/run/systemd/timesync/synchronized' does not exist, so use '/var/lib/systemd/timesync/clock'
+        if [ ! -e "${SYNCHRONIZED_FILE}" ]; then
+            SYNCHRONIZED_FILE="/var/lib/systemd/timesync/clock"
+        fi
+        # DynamicUser=yes moves the clock file to '/var/lib/private/systemd/timesync/clock'
+        if [ ! -e "${SYNCHRONIZED_FILE}" ]; then
+            SYNCHRONIZED_FILE="/var/lib/private/systemd/timesync/clock"
+        fi
         if [ -e "${SYNCHRONIZED_FILE}" ]; then
            FIND=$(( $(date +%s) - $(${STATBINARY} -L --format %Y "${SYNCHRONIZED_FILE}") ))
            # Check if last sync was more than 2048 seconds (= the default of systemd) ago


include/tests_time
```diff
@@ -574,7 +574,7 @@
 
 
     Register --test-no TIME-3185 --preqs-met "${PREQS_MET}" --weight L --network NO --category "security" --description "Check systemd-timesyncd synchronized time"
-    SYNCHRONIZED_FILE="/run/systemd/timesync/synchronized"
+    SYNCHRONIZED_FILE="/var/lib/systemd/timesync/clock"
```
konstruktoid, line 577: Maybe a if read test on `/run/systemd/timesync/synchronized` so we catch older versions as well?

Varbin: @konstruktoid `/var/lib/systemd/timesync/clock` exists in both old and new versions, but `/run/systemd/timesync/synchronized` only exists in new version (unfortunately I did not find an entry in the systemd changelog to tell which version it is), but I agree it would be better to test multiple files.

The systemd changelog also mentions ` /var/lib/private/systemd/timesync/clock`, but I am not sure under which circumstances this file exists: https://github.com/systemd/systemd/blob/master/NEWS#L2034-L2043. If I interpret it correctly, the file exists in `/var/lib/private` if DynamicUser option is used.

konstruktoid: You're correct, setting `DynamicUser=yes` in `/lib/systemd/system/systemd-timesyncd.service` creates `/var/lib/private/systemd/timesync/clock`.

```
$ sudo stat -c "%n: %z" /var/lib/systemd/timesync/clock /run/systemd/timesync/synchronized /var/lib/private/systemd/timesync/clock
/var/lib/systemd/timesync/clock: 2020-08-24 09:41:52.030011075 +0000
/run/systemd/timesync/synchronized: 2020-08-24 09:41:52.030011075 +0000
/var/lib/private/systemd/timesync/clock: 2020-08-24 09:41:52.030011075 +0000
$ systemd --version && uname -r && lsb_release -a
systemd 245 (245.4-4ubuntu3.2)
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid
5.4.0-42-generic
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.1 LTS
Release:        20.04
Codename:       focal
```

Jimver: @konstruktoid I added a check for both `/var/lib/systemd/timesync/clock` and `/var/lib/private/systemd/timesync/clock`

mboelen:
Thanks, merged!

PR #1016, opened 2020-08-26
ParseNginx(): Fix wildcard expansion, absolute path handling and output to stderr

### LINKED ISSUE #1015: ParseNginx() writes to stderr with certain nginx configurations
**Describe the bug**
The ParseNginx() function writes to stderr at the following line: https://github.com/CISOfy/lynis/blob/343e9bdc1c569f9a1590079e646c285542297e7e/include/functions#L2308

This happens if you have a line in your `nginx.conf` with an absolute path to a folder with a `*.conf` wildcard at the end and there are no matching files for that wildcard:
```
server {
        ...
        include /etc/nginx/conf.d/*.conf;
        ...
}
```

In addition, the `$VALUE` variable in:
https://github.com/CISOfy/lynis/blob/343e9bdc1c569f9a1590079e646c285542297e7e/include/functions#L2307
is already expanded while it should still contain the wildcard (as we are checking for the presence of the wildcard).

I am currently working on a PR to fix this (**Update**: PR is now open, see below)

**Version**
 - Distribution: Ubuntu 18.04
 - Lynis version: 3.0.1

**Expected behavior**
No output to stderr

**Output**

**Additional context**

Jimver:
This PR fixes #1015. 
- It changes the way the `$VALUE` variable is handled in the `ParseNginx()` function. The `$VALUE` variable is now not automatically expanded if it contains wildcards.
- In the `include` case of the nginx options now also properly processes absolute paths with wildcards
- ~~Pipe the stderr of the `ls` commands to `/dev/null` (happens in case the wildcard doesn't match anything)~~
- ~~Replace `ls` with `${LSBINARY}`~~
- Instead of using `ls`, use shell to resolve wildcard patterns


Varbin (changes requested):
I think `ls` should not be used here. If a configuration file contains special characters, `ls` will likely return something wrong (e.g all whitespaces are replaced by a space, if you have a recent copy of gnu ls, you will not see this behaviour). I know `ls` was already used in the original code (and in various other places).

See #826 for examples that can be problematic. While some of the examples I used are quite unlikely, some others (e.g. paths with spaces) are not, and as far I know nginx supports quoting paths to include spaces (and *.conf should match files with spaces). 

As far I know, the `ls` could be skipped, by iterating over the unqoted wildcard, as the shell will automatically resolve this (e.g. `for FOUND_CONF in "${VALUE}/"${WILDCARD}; do`).

Jimver:
Good point, didn't know that about `ls`! 

One question: If we let it resolve by the shell we should explicitly check if it matches anything right? Otherwise it will remain the wildcard string I assume

Varbin:
> One question: If we let it resolve by the shell we should explicitly check if it matches anything right? Otherwise it will remain the wildcard string I assume

Yes.

```bash
$ for i in *.nonexistent; do echo "$i"; done
*.nonexistent
```

Jimver:
@Varbin Alright I now removed the `ls` command and use shell to resolve the wildcards!

mboelen:
Thanks and merged!

Issue #1014, opened 2020-08-24
FlatCar OS not recognised


pvillega:
**Describe the bug**
Lynis doesn't detect Flatcar on OS detection

**Version**
 - FlatCar OS 4.19.140-flatcar
 - Lynis version 3.0.1

**Expected behavior**
Lynis detects Flatcar as the OS

**Output**
```
  Exception found!

  Function/test:  [OS Detection]
  Message:        Unknown OS found in /etc/os-release - Please create issue on GitHub project page: https://github.com/CISOfy/lynis

  Help improving the Lynis community with your feedback!
```

Varbin:
What is the content of `/etc/os-release`?

pvillega:
Mine shows:

```
core@hght ~ $ cat /etc/os-release
NAME="Flatcar Container Linux by Kinvolk"
ID=flatcar
ID_LIKE=coreos
VERSION=2512.3.0
VERSION_ID=2512.3.0
BUILD_ID=2020-08-19-1355
PRETTY_NAME="Flatcar Container Linux by Kinvolk 2512.3.0 (Oklo)"
ANSI_COLOR="38;5;75"
HOME_URL="https://flatcar-linux.org/"
BUG_REPORT_URL="https://issues.flatcar-linux.org"
FLATCAR_BOARD="amd64-usr"
```

PR #1017, opened 2020-08-27
Add support for Flatcar Container Linux

Varbin:
Fixes cisofy/lynis#1014.

Flatcar is a fork of CoreOS. Thus the variable LINUX_VERSION_LIKE
(introduced with #1004) for Flatcar is CoreOS.


mboelen:
Thanks, merged!

PR #1029, opened 2020-10-11
Add test for Suricata IDS/IPS

silentcreek:
Commit 94e0a4e added a test for the Suricata binary, but the result appears to be used nowhere. Add a proper test for an active Suricata daemon in the IDS/IPS tooling section.

I avoided additional tests to verify the Suricata configuration here because Suricata's test mode involves compiling the entire detection ruleset which can take several minutes depending on the configuration, size of the ruleset and system.


mboelen:
Great work, it has been merged!

Issue #1035, opened 2020-10-17
Runtime error


steph78630:
**Describe the bug**
When running in developer mode, displays errors messages.

**Version**
 - Distribution [IPfire core update 151 dev]
 - Lynis version [3.0.0.2 up to date]

**Expected behavior**
No error message displayed.

**Output**

**Additional context**
./lynis/include/tests_boot_services: line 996: SYSTEMDANALYZEBINARY: unbound variable
./lynis/include/tests_nameservices: line 176: DNSDOMAINNAMEBINARY: unbound variable
./lynis/include/tests_networking: line 245: RESOLVECTLBINARY: unbound variable
./lynis/include/tests_time: line 84: NTPCTLBINARY: unbound variable
./lynis/include/tests_time: line 231: TIMEDATECTL: unbound variable
./lynis/include/tests_crypto: line 152: CRYPTSETUPBINARY: unbound variable
./lynis/include/tests_file_integrity: line 303: EVMCTLBINARY: unbound variable
./lynis/include/tests_file_integrity: line 325: INTEGRITYSETUPBINARY: unbound variable
./lynis/include/tests_file_integrity: line 365: VERITYSETUPBINARY: unbound variable

Edit (add new missing constants) :
./lynis/plugins/plugin_pam_phase1: line 249: PAM_PASSWORD_PWHISTORY_AMOUNT: unbound variable
./lynis/include/tests_networking: line 470: SSBINARY: unbound variable
./lynis/include/tests_crypto: line 152: LSBLKBINARY: unbound variable

That's all !

PR #1039, opened 2020-10-18
Add missing constants

steph78630:
When running tests in debug mode, messages indicate some missing constants (see issue #1035)


steph78630:
Delivery of 2 files in error

Issue #1034, opened 2020-10-17
Add 'Mageia' in osdetection module


rexy74:
**Describe the bug**
Unknown OS found in /etc/os-release

**Version**
 - Distribution : Mageia-7.1
 - Lynis version : 3.0.1

**Expected behavior**
mach the content of "/etc/os-release" file

**Output**
Unknown OS found in /etc/os-release

**Additional context**
This is the content of os-release :
NAME="Mageia"
VERSION="7"
ID=mageia
VERSION_ID=7
ID_LIKE="mandriva fedora"
PRETTY_NAME="Mageia 7"
ANSI_COLOR="1;36"
HOME_URL="http://www.mageia.org/"
SUPPORT_URL="http://www.mageia.org/support/"
BUG_REPORT_URL="https://bugs.mageia.org/"
PRIVACY_POLICY_URL="https://wiki.mageia.org/en/Privacy_policy"

konstruktoid:
Hi @rexy74, Mageia was added a few years back (https://github.com/CISOfy/lynis/blame/master/include/osdetection#L399-L404) can you check if `grep "^Mageia" /etc/redhat-release | awk '{ if ($2=="release") { print $3 } }'` fails?

konstruktoid:
Mageia got a `/etc/os-release` and will get caught at https://github.com/CISOfy/lynis/blob/master/include/osdetection#L142 and not at https://github.com/CISOfy/lynis/blob/master/include/osdetection#L373.
Can you test https://github.com/CISOfy/lynis/pull/1043?

rexy74:
Hi @konstruktoid, this is all you need to know which file to take into account :
------
ls -l /etc/*-release
-rw-r--r-- 1 root root 104 sept. 20  2018 lsb-release
-rw-r--r-- 1 root root  39 juil. 12  2019 mageia-release
lrwxrwxrwx 1 root root  14 juil. 12  2019 mandrakelinux-release -> mageia-release
lrwxrwxrwx 1 root root  14 juil. 12  2019 mandrake-release -> mageia-release
lrwxrwxrwx 1 root root  14 juil. 12  2019 mandriva-release -> mageia-release
lrwxrwxrwx 1 root root  21 juil. 12  2019 os-release -> ../usr/lib/os-release
lrwxrwxrwx 1 root root  14 juil. 12  2019 redhat-release -> mageia-release
lrwxrwxrwx 1 root root  14 juil. 12  2019 system-release -> mageia-release
------
[...]

PR #1043, opened 2020-10-19
Mageia got /etc/os-release and add Mageia EOL dates

konstruktoid:
Add Mageia EOL dates and `/etc/redhat-release` is a link to `/etc/mageia-release`.

Ref https://github.com/CISOfy/lynis/issues/1034


konstruktoid:
Mageia got a `/etc/os-release` and will get caught at https://github.com/CISOfy/lynis/blob/master/include/osdetection#L142 and not at https://github.com/CISOfy/lynis/blob/master/include/osdetection#L373.

mboelen:
Thanks, merged!

PR #1058, opened 2020-10-25
Use X/Open grep, nawk and GNU echo on Solaris

Varbin:
On Solaris the default grep does not support `-E` and `-q`.  The first issue is fixed by using X/Open grep from `/usr/xpg4/bin/grep` and the latter by using shell redirections instead.

`awk` is replaced by `nawk` on Solaris, too, as only the latter has the required feature set.

### DIFF
--- a/include/binaries
+++ b/include/binaries
@@ -233,7 +233,7 @@
                             osiris)                 OSIRISBINARY="${BINARY}";          LogText "  Found known binary: osiris - ${BINARY}" ;;
                             openssl)                OPENSSLBINARY="${BINARY}";         OPENSSLVERSION=$(${BINARY} version 2> /dev/null | head -n 1 | awk '{ print $2 }' | xargs); LogText "Found ${BINARY} (version ${OPENSSLVERSION})" ;;
                             pacman)
-                                if [ -z "$(echo "${BINARY}" | grep -E "/usr(/local)?/games")" ]; then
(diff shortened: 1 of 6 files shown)


mboelen (changes requested):
The `od` is there as a protection, so should not be removed. The command not available?

Varbin:
@mboelen Which `od` do you mean, I can't find it in the changes of this PR? I only remember removing one in #1069.

mboelen:
This specific one: https://github.com/CISOfy/lynis/pull/1058/commits/05addd0bdf684c755dae15ab7f8a35b52dfc0368

Varbin:
> This specific one: [05addd0](https://github.com/CISOfy/lynis/commit/05addd0bdf684c755dae15ab7f8a35b52dfc0368)

@mboelen The contents of this commit are already in the master branch. It has been misplaced be me, feel free remove it.
To verify you can run a diff between this PR and the current master (or see the files changed tab on Github).

The removal of `od` at this line already happened with my PR 1069. `od` is still present before displaying the found data (for some reasons it is not part of the commit on this PR). It must be used directly before echoing, as it will insert spaces on some platforms.

mboelen:
This pull request is open for a while, mainly because it can't be easily merged (several changes). With an additional merge conflict (not caused by you), it becomes even harder.

As part of a cleanup routine, I will close this pull request at this moment. If you have the chance, please create several smaller pull requests (one change per PR). Then we can review and merge them.

PR #1007, opened 2020-08-09
Fix multiple issues on Solaris 11

Varbin:
Fixes #1005. Below is a list of of changes and solved problems:
 
 - Issues with grep and egrep are fixed
 - Use GNU echo for echo -n or printf and nawk instead of awk
 - Include support for Solaris netstat: 
 - Fix check for existence of grub.d
 - Do hostname checking in two steps, as Solaris' tr does not support full regular expressions
 - Test if loghost is not localhost and remote logging to remote localhost loghost is not detected as remote logging (LOGG-2153)
 - Support for Solaris' 11 package manager:
   -  `pkg audit` is not falsely detected as package audit software
   - Support repository installation is tested (equivalent to security repository on Debian, but one has to buy access)
   - Detect solaris-11-cpu as audit tool (vulnerability database for IPS / pkg), it requires the to the support repository
   - Other tests as package enumeration, available updates, ...
   - New tests: PKGS-7338, PKGS-7339, PKGS-7340, PKGS-7341
 - Detect sshd at the uncommon Solaris location and fix the awk operation not supported on Solaris, so the configuration check was skipped.
 - Detect IPS service manager (or SysV) in  BOOT-5170 (new). The SysV permission test BOOT-5184 is executed on Solaris.
 - Detect Trusted Solaris as MAC framework (new tests MACF-6250 and MACF-6252). Linux MAC and RBAC framework tests are now only executed on Linux.

Open issues:
 - Ownership problems: Solaris uses different permissions than most other Unix systems. The "classic" bin, sys, ... groups are still used and and can be checked with pkg verify. I don't think this can only solved with support for os specific permission tests.

PR #1059, opened 2020-10-25
Test if /etc/grub.d is a directory

Varbin:
Part of the Solaris fixes of #1007, but is likely relevant for other operating systems as well.


mboelen:
Good catch and makes sense. Merged!

PR #1061, opened 2020-10-25
Simplify tr hostname checking expression

Varbin:
Solaris' tr does not support full regular expressions.


mboelen (changes requested):
:alpha: should be :alnum: so it includes 0-9

Does `tr -d '[:alnum:]\.\-'` work correctly on Solaris?

mboelen:
Thanks, merged!

PR #1062, opened 2020-10-25
Test if loghost is not localhost

Varbin:
On Solaris, the name loghost can be used to point to remote log servers.
By default loghost is configured to 127.0.0.1, logging to the local
machine.

Thus a new test - LOGG-2153 - is created to test if loghost is not
localhost and LOGG-2154 is modified to ignore @loghost lines if loghost
is localhost.


mboelen:
Thanks, merged!

PR #1064, opened 2020-10-25
Add support for Solaris services, run BOOT-5184 there

Varbin:
The Solaris ~~IPS~~ SFM service manager (svcs) is now detected, and services
managed with it are enumerated.

Test BOOT-5184 now runs on Solaris, too, as SysV init scripts are
supported as well, even with ~~IPS~~ SFM. SysV Init has been the traditional
init system on Solaris and is used in previous versions.


Varbin:
The service manager is not named IPS, but SFM. I adjusted the PR accordingly.

mboelen (approved):

mboelen:
Merged with small change. Thanks!

Issue #1065, opened 2020-10-26
Errors during profile check on OpenSolaris distributions


Varbin:
**Describe the bug**
On OpenSolaris derivates (e.g. Omnios and OpenIndiana), Lynis does not start and reports "unexpected characters in profile".

**Version**
 - OmniosCE r151034l / OpenIndiana 
 - Lynis version 3.0.2 (499cf1c or Rev 3182 if copied as SVN repository)

**Expected behavior**
Lynis starts.

**Output**

OpenIndiana (had to clone as svn repository):
<pre>
root@openindiana:~/lynis.git/trunk# ./lynis audit system

[ Lynis 3.0.2 ]

################################################################################
  Lynis comes with ABSOLUTELY NO WARRANTY. This is free software, and you are
  welcome to redistribute it under the terms of the GNU General Public License.
  See the LICENSE file for details about using this software.

  2007-2020, CISOfy - https://cisofy.com/lynis/
  Enterprise support available (compliance, plugins, interface and tools)
################################################################################

[+] Initializing program
------------------------------------
  - Detecting OS...                                           [ DONE ]

 ==================================================================================================

  [WARNING]: Your profile contains old-style configuration entries. See log file for more details and how to convert these entries

  * ISSUE
  Your profile has one or more lines that are in an old format (key:value). They need to be converted into the new format (key=value) or disabled.

  * HOW TO RESOLVE
  Use grep to see the relevant matches (grep -E "^[a-z-]{1,}:" custom.prf)

 ==================================================================================================

  [WARNING]: Your profile '/root/lynis.git/trunk/default.prf' contains unexpected characters. See the log file for more information.

Fatal error: unexpected characters in profile. Stopping execution (security measure)
</pre>

OmniosCE:
<pre>
root@omniosce:~/lynis# ./lynis audit system

[ Lynis 3.0.2 ]

################################################################################
  Lynis comes with ABSOLUTELY NO WARRANTY. This is free software, and you are
  welcome to redistribute it under the terms of the GNU General Public License.
  See the LICENSE file for details about using this software.

  2007-2020, CISOfy - https://cisofy.com/lynis/
  Enterprise support available (compliance, plugins, interface and tools)
################################################################################

[+] Initializing program
------------------------------------
awk: syntax error near line 1
awk: bailing out near line 1
  - Detecting OS...                                           [ DONE ]

 ==================================================================================================

  [WARNING]: Your profile contains old-style configuration entries. See log file for more details and how to convert these entries

[...]

PR #1069, opened 2020-10-30
Use `od` only for displaying invalid characters

Varbin:
The first od is removed, the second time is moved to right before echoing
the characters. On certain OpenSolaris distributions, `od` always outputs
spaces, even if the input is empty. The spaces would have been converted
to !space!, thus Lynis detected invalid characters / old style configuration.

Resolves cisofy/lynis#1065.

Signed-off-by: Simon Biewald <simon@fam-biewald.de>

### DIFF
--- a/include/profiles
+++ b/include/profiles
@@ -35,7 +35,7 @@
 
         # Show deprecation message for old config entries such as 'config:' and 'apache:'
         FOUND=0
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, merged!

PR #1071, opened 2020-11-09
OpenSolaris distribution detection

Varbin:
This PR adds detection of OpenSolaris distributions. The variable OPENSOLARIS is introduced to distringuish between the closed  source Solaris or OpenSolaris based ones. The edge case of OpenSolaris itself is not yet solved, but OpenSolaris itself should be very rare these days. Solaris Express is not detected yet as well.

Lynis will try to read the `/etc/os-release` if possible and will fall back to parsing the free form `/etc/release` file.
Currently detected and distinguished (Open)Solaris variants with correct version number are:
 - Solaris >= 11 (exluding Solaris Express and OpenSolaris) as "Oracle Solaris" with EOL dates for 11.3 and .4
 - Solaris < 11 as "Sun Solaris" 
 - OmniosCE (but not old Omnios) with EOL dates
 - OpenIndiana (Hipster and Legacy)
 - Shillix
 - SmartOS
 - Tribblix
 - "Unknown Illumos" for unknown distributions based on Illumos

Lynis will fall back to "Sun Solaris" with "SunOS 5.X" as full name for unknown distributions (this is the current behaviour), if both parsing the `os-release` and `release` files do not return meaningful results.

### DIFF
--- a/db/software-eol.db
+++ b/db/software-eol.db
@@ -237,4 +237,20 @@ os:Ubuntu 18.10:2019-07-18:1563400800:
 os:Ubuntu 19.04:2020-01-01:1577833200:
 os:Ubuntu 20.04:2025-04-01:1743458400:
 #
+# OmniosCE - https://omniosce.org/releasenotes.html
+#
+os:OmniOS Community Edition v11 r151022:2020-05-11:1589148000:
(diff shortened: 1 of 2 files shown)


mboelen:
Awesome! Merged :)

Issue #1075, opened 2020-11-14
HostID generation fails on OmniosCE (and likely other Solaris based)


Varbin:
**Describe the bug**

Lynis generates an exception with the message "No interface found op Solaris to create HostID".

**Version**
 - OmniOS Community Edition r151034l
 - Lynis version 3.0.2 with my (Open)Solaris patches applied

**Expected behavior**
No exceptions 😄

**Output**

```
[ Lynis 3.0.2 ]

################################################################################
  Lynis comes with ABSOLUTELY NO WARRANTY. This is free software, and you are
  welcome to redistribute it under the terms of the GNU General Public License.
  See the LICENSE file for details about using this software.

  2007-2020, CISOfy - https://cisofy.com/lynis/
  Enterprise support available (compliance, plugins, interface and tools)
################################################################################


[+] Initializing program
------------------------------------
  - Detecting OS...                                           [ DONE ]
  - Checking profiles...                                      [ DONE ]

  ---------------------------------------------------
  Program version:           3.0.2
  Operating system:          Solaris
  Operating system name:     OmniOS
  Operating system version:  r151034l
  Operating system mode:     64
  Kernel version:            omnios-r151034-831ff8e83b
  Hardware platform:         i86pc
  Hostname:                  omniosce
  ---------------------------------------------------
  Profiles:                  /root/lynis-3/default.prf
  Log file:                  /var/log/lynis.log
  Report file:               /var/log/lynis-report.dat
  Report version:            1.0
  Plugin directory:          ./plugins
  ---------------------------------------------------
  Auditor:                   [Not Specified]
  Language:                  en
  Test category:             all
  Test group:                all
  ---------------------------------------------------
  - Program update status...                                  [ NO UPDATE ]

[+] System tools
------------------------------------
  - Scanning available tools...
  - Checking system binaries...

[+] Plugins (phase 1)
------------------------------------
 Note: plugins have more extensive tests and may take several minutes to complete

  - Plugin: pam
    [..]
  - Plugin: systemd
    [................]

=================================================================

  Exception found!

  Function/test:  [GetHostID]
  Message:        No interface found op Solaris to create HostID

  Help improving the Lynis community with your feedback!

  Steps:
  - Ensure you are running the latest version (./lynis update check)
  - If so, create a GitHub issue at https://github.com/CISOfy/lynis
  - Include relevant parts of the log file or configuration file

  Thanks!

=================================================================

(Snip)
```

`lynis show hostid`:

```
[...]

PR #1076, opened 2020-11-14
Use first e1000 interface and break after match (for hostid generation on Solaris)

Varbin:
Fixes CISOfy/lynis#1075.

Before this commit, the interfaces "e1000g1" and "net0" were allowed. The name "e1000g0" is appended to the list. After finding an interface, the loop is interrupted now. As previously "net0" was always used, even if another interface was available, the list is reordered to "net0 e1000g1 e1000g0" to not break previous generations.

A typo is also fixed ("No interface found op Solaris ..." -> "No interface found on Solaris ...").

Signed-off-by: Simon Biewald <simon@fam-biewald.de>

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -1089,12 +1089,13 @@
                 ;;
 
                 "Solaris")
-                    INTERFACES_TO_TEST="e1000g1 net0"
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks, merged!

PR #1087, opened 2020-12-14
only fail AUTH-9230 if both SHA_CRYPT_{MIN,MAX}_ROUNDS are undefined

### LINKED ISSUE #1085: Incorrect interpretation of SHA_CRYPT_{MIN,MAX}_ROUNDS in AUTH-9230
**Describe the bug**
Lynis test AUTH-9230 complains that `SHA_CRYPT_{MIN,MAX}_ROUNDS` is not sufficiently hardened when only `SHA_CRYPT_MIN_ROUNDS` is configured in `/etc/login.defs`.

`login.defs` states the following:

```
#
# Only works if ENCRYPT_METHOD is set to SHA256 or SHA512.
#
# Define the number of SHA rounds.
# With a lot of rounds, it is more difficult to brute forcing the password.
# But note also that it more CPU resources will be needed to authenticate
# users.
#
# If not specified, the libc will choose the default number of rounds (5000).
# The values must be inside the 1000-999999999 range.
# If only one of the MIN or MAX values is set, then this value will be used.
# If MIN > MAX, the highest value will be used.
#
```

"If only one of the MIN or MAX values is set, then this value will be used.", so configuring only `SHA_CRYPT_MIN_ROUNDS` should be enough.

**Version**
 - Distribution Slackware64 current
 - Lynis version 3.0.1

**Expected behavior**
Lynis should be happy when only `SHA_CRYPT_MIN_ROUNDS` is configured

**Output**
```
2020-12-13 20:11:47 Performing test ID AUTH-9230 (Check group password hashing rounds)
2020-12-13 20:11:47 Test: Checking SHA_CRYPT_MIN_ROUNDS option in /etc/login.defs
2020-12-13 20:11:47 Result: number of encryption algorithm rounds is 500000
2020-12-13 20:11:47 Hardening: assigned maximum number of hardening points for this item (2). Currently having 14 points (out of 22)
2020-12-13 20:11:47 Test: Checking SHA_CRYPT_MAX_ROUNDS option in /etc/login.defs
2020-12-13 20:11:47 Result: number of maximum rounds used by the encryption algorithm is not configured
2020-12-13 20:11:47 Suggestion: Configure maximum encryption algorithm rounds in /etc/login.defs [test:AUTH-9230] [details:-] [solution:-]
2020-12-13 20:11:47 Hardening: assigned partial number of hardening points (0 of 2). Currently having 14 points (out of 24)
2020-12-13 20:11:47 ====
```

konstruktoid:
Since if only one of the MIN or MAX values is set, and then this value will be used, we only have to fail AUTH-9230 if both SHA_CRYPT_{MIN,MAX}_ROUNDS are undefined.

Closes #1085 

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -352,50 +352,49 @@
 #################################################################################
 #
     # Test        : AUTH-9230
-    # Description : Check group password hashing rounds in login.defs
+    # Description : Check password hashing rounds in login.defs
     # Notes       : Applicable to all Unix-like OS
     PREQS_MET="NO"
     if [ -f ${ROOTDIR}etc/login.defs ]; then
         PREQS_MET="YES"
     fi
-    Register --test-no AUTH-9230 --preqs-met ${PREQS_MET} --root-only NO --weight L --network NO --category security --description "Check group password hashing rounds"
+
+    Register --test-no AUTH-9230 --preqs-met ${PREQS_MET} --root-only NO --weight L --network NO --category security --description "Check password hashing rounds"
     if [ ${SKIPTEST} -eq 0 ]; then
         SHA_CRYPT_MIN_ROUNDS_FIND=$(${GREPBINARY} "^SHA_CRYPT_MIN_ROUNDS" ${ROOTDIR}etc/login.defs | ${AWKBINARY} '{ if ($1=="SHA_CRYPT_MIN_ROUNDS") { print $2 } }')
         SHA_CRYPT_MAX_ROUNDS_FIND=$(${GREPBINARY} "^SHA_CRYPT_MAX_ROUNDS" ${ROOTDIR}etc/login.defs | ${AWKBINARY} '{ if ($1=="SHA_CRYPT_MAX_ROUNDS") { print $2 } }')
+        SHA_CRYPT_ROUNDS=0
 
-        if [ -z "${SHA_CRYPT_MIN_ROUNDS_FIND}" -o "${SHA_CRYPT_MIN_ROUNDS_FIND}" = "0" ] && [ -z "${SHA_CRYPT_MAX_ROUNDS_FIND}" -o "${SHA_CRYPT_MAX_ROUNDS_FIND}" = "0" ]; then
-            LogText "Result: number of rounds used by the encryption algorithm is not configured"
-            Display --indent 2 --text "- Checking group password hashing rounds" --result "${STATUS_DISABLED}" --color YELLOW
-            ReportSuggestion "${TEST_NO}" "Configure encryption algorithm rounds in /etc/login.defs"
+        if [ -n "${SHA_CRYPT_MIN_ROUNDS_FIND}" -a -n "${SHA_CRYPT_MAX_ROUNDS_FIND}" ]; then
+            if [ ${SHA_CRYPT_MIN_ROUNDS_FIND} -lt ${SHA_CRYPT_MAX_ROUNDS_FIND} ]; then
+                SHA_CRYPT_ROUNDS=${SHA_CRYPT_MIN_ROUNDS_FIND}
+            else
+              SHA_CRYPT_ROUNDS=${SHA_CRYPT_MAX_ROUNDS_FIND}
+            fi
+        elif [ -z "${SHA_CRYPT_MIN_ROUNDS_FIND}" -a -n "${SHA_CRYPT_MAX_ROUNDS_FIND}" ]; then
+            SHA_CRYPT_ROUNDS=${SHA_CRYPT_MAX_ROUNDS_FIND}
+        elif [ -n "${SHA_CRYPT_MIN_ROUNDS_FIND}" -a -z "${SHA_CRYPT_MAX_ROUNDS_FIND}" ]; then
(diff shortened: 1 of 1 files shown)


pyllyukko:
I think the test should be written like the description in `login.defs` (with `if ... elif`):

* If `MIN` OR `MAX` is defined we'll use whichever is defined
* If both are defined and `MIN` < `MAX` -> we'll use `MIN`
* If both are defined and `MIN` > `MAX` -> use `MIN`

konstruktoid:
It now verifies the lowest or only available option.

Even though`MIN > MAX, the highest value will be used`, setting a `MAX` value lower than `MIN` should result in a suggestion since it's counterintuitive in my opinion.

include/tests_authentication
```diff
@@ -352,50 +352,48 @@
...
-                AddHP 1 2
+        if [ -n "${SHA_CRYPT_MIN_ROUNDS_FIND}" -a -n "${SHA_CRYPT_MAX_ROUNDS_FIND}" ]; then
+            if [ ${SHA_CRYPT_MIN_ROUNDS_FIND} -lt ${SHA_CRYPT_MAX_ROUNDS_FIND} ]; then
+                SHA_CRYPT_ROUNDS=${SHA_CRYPT_MIN_ROUNDS_FIND}
             else
-                LogText "Result: number of encryption algorithm rounds is ${SHA_CRYPT_MIN_ROUNDS_FIND}"
-                PASSWORD_MINIMUM_ROUNDS=${SHA_CRYPT_MIN_ROUNDS_FIND}
-                Display --indent 2 --text "- Group password hashing rounds (minimum)" --result CONFIGURED --color GREEN
-                AddHP 2 2
+              SHA_CRYPT_ROUNDS=${SHA_CRYPT_MAX_ROUNDS_FIND}
```
pyllyukko, line 371: This is indented with two spaces instead of four.

pyllyukko:
> Even though`MIN > MAX, the highest value will be used`, setting a `MAX` value lower than `MIN` should result in a suggestion since it's counterintuitive in my opinion.

I agree.

include/tests_authentication
```diff
@@ -352,50 +352,50 @@
...
         fi
 
-        LogText "Test: Checking SHA_CRYPT_MAX_ROUNDS option in ${ROOTDIR}etc/login.defs"
-        FIND=$(${GREPBINARY} "^SHA_CRYPT_MAX_ROUNDS" ${ROOTDIR}etc/login.defs | ${AWKBINARY} '{ if ($1=="SHA_CRYPT_MAX_ROUNDS") { print $2 } }')
-        if [ -z "${FIND}" -o "${FIND}" = "0" ]; then
-            LogText "Result: number of maximum rounds used by the encryption algorithm is not configured"
-            Display --indent 2 --text "- Checking maximum group password hashing rounds" --result "${STATUS_DISABLED}" --color YELLOW
-            ReportSuggestion "${TEST_NO}" "Configure maximum encryption algorithm rounds in /etc/login.defs"
+        LogText "Test: Checking SHA_CRYPT_{MIN,MAX}_ROUNDS option in ${ROOTDIR}etc/login.defs"
+        if [ ${SHA_CRYPT_ROUNDS} -eq 0 ]; then
```
pyllyukko, line 382: This comparison breaks if `SHA_CRYPT_ROUNDS` is `potato`. The original one also, but just wondering if there should be safeguards to check whether it's an integer at all.

include/tests_authentication
```diff
@@ -352,50 +352,50 @@
...
-            Display --indent 2 --text "- Group password hashing rounds (maximum)"  --result "${STATUS_SUGGESTION}" --color YELLOW
-            AddHP 1 2
-        else
-            LogText "Result: number of encryption algorithm rounds is ${FIND}"
-            PASSWORD_MINIMUM_ROUNDS=${FIND}
-            Display --indent 2 --text "- Group password hashing rounds (maximum)" --result CONFIGURED --color GREEN
-            AddHP 2 2
+        fi
+
+        if [ -n "${SHA_CRYPT_ROUNDS}" ] && [ ${SHA_CRYPT_ROUNDS} -gt 0 ]; then
```
pyllyukko, line 389: Isn't this comparison redundant?

include/tests_authentication
pyllyukko: Maybe use [DigitsOnly()](https://github.com/CISOfy/lynis/blob/701be9852bae4f64adf7b94f874956ffe1fffb54/include/functions#L403-L417) for `SHA_CRYPT_ROUNDS`?

include/tests_authentication
konstruktoid: Who indents with four when it's not yaml...

include/tests_authentication
konstruktoid: I set `SHA_CRYPT_ROUNDS = 0` when the value is empty, https://github.com/CISOfy/lynis/blob/807cfd430a434331ad108677f51bc1586de236ec/include/tests_authentication#L382

include/tests_authentication
konstruktoid: Seems reasonably, but `DigitsOnly()` is only used in `plugins/plugin_pam_phase1` so that should probably be another PR where all values are validated.

mboelen:
Thanks, merged!

PR #1097, opened 2021-01-01
Krnl 5677

bginsbach:
Allow for multiple hardware names in Register. Primarily added to support *BSD systems that use amd64 as the hardware name rather than the Linux (and Intel) x86_64. Additionally this allows for a test to possibly support multiple hardware types.

Open to discussing alternatives rather than just letting this PR timeout this time.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -2670,7 +2670,15 @@
         fi
 
         # Check for correct hardware platform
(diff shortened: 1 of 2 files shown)


mboelen:
Thanks, merged!

PR #1099, opened 2021-01-03
fix syntax must include ${} 

gcsgithub:


mboelen:
Thanks, good catch! :)

gcsgithub:
I can only guess its an old KSH programmer as typeset -i  would have made that valid syntax

PR #1102, opened 2021-01-03
New test: BOOT-5140 - Check for ELILO boot loader presence

teoberi:
Add TestID for ELILO
Add test for ELILO boot loader


teoberi:
?

PR #1128, opened 2021-03-31
nftables empy ruleset test fix by reorder --stateless

nberlee:
```
$ nft -v
nftables v0.9.8 (E.D.S.)
$ sudo nft list ruleset --stateless 
Error: syntax error, options must be specified before commands
```


mboelen:
Merged, thanks!

PR #1137, opened 2021-04-23
Lynis with PostgreSQL 13.1 test ID DBS-1826/DBS-1828

teoberi:
Fix the variable name for PostgreSQL
Fix for Test: DBS-1826 - Check if PostgreSQL is being used
Fix for Test: DBS-1828 - Test PostgreSQL configuration file(s)


teoberi:
Without correcting the process name from "postgres:" to "postgres" detection for PostgreSQL does not work.
Also when compiling from source the installation path is: "usr/local/pgsql/data".

teoberi:
This is the second time I have added this pull request.
If there is something wrong, it can be corrected or I can give up!

mboelen:
Thanks for your work on this, it has been merged. Also my apologies for the delay, we had a bigger batch of work. When things get stalled next time, feel free to poke me directly by email.

PR #1143, opened 2021-05-04
[LOGG-2138] Do not check for klogd when metalog is being used

jhe2:
Metalog comes with its own kernel logger, so there is no need to also run klogd.


jhe2:
I've changed the expression of the ``if`` statement to not use the ``-a`` flag of the ``[`` command, since its use is discouraged by POSIX, in favor of individual ``[`` commands in conjunction with the ``&&`` operator. This is also very commonly used in other places of Lynis, so I hope you are okay with it.

PR #1169, opened 2021-06-15
Fix Big Sur Detection and add macOS Monterey

al-lac:
Fixed the macOS Big Sur (Version 11) detection and added support for macOS Monterey (Version 12)


mboelen:
Thanks, merged!

PR #1170, opened 2021-06-15
Update german language (de)

al-lac:
Add missing german strings


mboelen:
Thanks! It has been merged

Issue #1179, opened 2021-06-30
Unknown OS - Artix Linux


digitalcheetah:
**Describe the bug**
An unknown OS exception is thrown when running on Artix Linux.

**Version**
 - Distribution: Artix Linux
 - Lynis version: 3.0.4

**Expected behavior**
It should be detected.

**Output**
2021-07-01 00:39:08 EOL check: 255
2021-07-01 00:39:08 Note: the end-of-life of 'Artix Linux' could not be checked. Entry missing in software-eol.db?

**Additional context**
From `/etc/os-release`:

NAME="Artix Linux"
PRETTY_NAME="Artix Linux"
ID=artix
BUILD_ID=rolling
ANSI_COLOR="0;36"
HOME_URL="https://www.artixlinux.org/"
DOCUMENTATION_URL="https://wiki.artixlinux.org/"
SUPPORT_URL="https://forum.artixlinux.org/"
BUG_REPORT_URL="https://bugs.artixlinux.org/"
LOGO=artixlinux

Artix Linux, like Arch, is a rolling distribution.

mboelen:
Thanks for reporting!

Do you have the chance to add an entry to include/os-detection and become a contributor? :)

digitalcheetah:
Sure. Just made the PR a few minutes ago. :)

mboelen:
Closing now pull request has been merged

PR #1188, opened 2021-07-11
Adding Artix Linux to osdetection

digitalcheetah:
By request of #1179, in order to add Artix Linux to lynis.


mboelen:
Thanks, merged!

Issue #1167, opened 2021-06-14
include/tests_networking, change grep --files-with-matches --no-messages to -l -s


zcrayfish:
On systems using non-GNU versions of grep, such as busybox grep, the GNU long options are not as well supported and will result in errors being printed.

**Version**
 - Distribution: Alpine Linux v3.13.5
 - Lynis version: 3.0.5

**Expected behavior**
binaries called by the lynis script should not use GNU long options.

**Output**
```
/bin/grep: unrecognized option: files-with-matches
BusyBox v1.32.1 () multi-call binary.

Usage: grep [-HhnlLoqvsrRiwFE] [-m N] [-A/B/C N] PATTERN/-e PATTERN.../-f FILE [FILE]...

Search for PATTERN in FILEs (or stdin)

        -H      Add 'filename:' prefix
        -h      Do not add 'filename:' prefix
        -n      Add 'line_no:' prefix
        -l      Show only names of files that match
        -L      Show only names of files that don't match
        -c      Show only count of matching lines
        -o      Show only the matching part of line
        -q      Quiet. Return 0 if PATTERN is found, 1 otherwise
        -v      Select non-matching lines
        -s      Suppress open and read errors
        -r      Recurse
        -R      Recurse and dereference symlinks
        -i      Ignore case
        -w      Match whole words only
        -x      Match whole lines only
        -F      PATTERN is a literal (not regexp)
        -E      PATTERN is an extended regexp
        -m N    Match up to N times per file
        -A N    Print N lines of trailing context
        -B N    Print N lines of leading context
        -C N    Same as '-A N -B N'
        -e PTRN Pattern to match
        -f FILE Read pattern from file
```

**Additional context**
According to the GNU grep manpage, the --files-with matches long option has a POSIX equivalent: -l
Likewise the --no-messages long option has a POSIX equivalent of: -s

PR #1194, opened 2021-07-24
grep options change from gnu to posix

zcrayfish:
fix issue #1167


mboelen:
Thanks, merged! Added a comment, due to using short notation which may not directly clear what each option does.

PR #1204, opened 2021-09-01
Update EOL for Alpine Linux release

kolenichsj:
Added EOL date for Alpine Linux 3.14


kolenichsj:
Added EOL date for Alpine Linux 3.15

db/software-eol.db
```diff
@@ -16,6 +16,8 @@
 #
 # Alpine - https://alpinelinux.org/releases/
 #
+os:Alpine 3.15:2023-11-01:1169879680
```
Jaimie85, line 19: There seems to be an error in the converted time

Jaimie85 (changes requested):
There seems to be an error in the converted time.

db/software-eol.db
mboelen: Fixed with my change

kolenichsj:
Thank you for the correction

PR #1205, opened 2021-09-01
Adding support for Alpine Package Keeper

kolenichsj:
- Added known binary apk
- Updated function 'PackageIsInstalled' to use apk when available
- Added two tests to tests_ports_packages
  - PKGS-7346 - Check apk and log/report installed packages
  - PKGS-7395 - Check apk for packages that can be upgraded


initpwn (approved):

mboelen:
Thanks, looks good. Merged!

PR #1215, opened 2021-10-13
tests_filesystems: check for blacklisted modules also under /usr/lib/modules.d

jsegitz:
On newer (open)SUSE systems and probably also other systems that move distribution config files to /usr the files live there


jsegitz:
just notice that this is still open. Anything blocking a merge?

jsegitz:
wrong button

mboelen:
Sorry it took a while, it had a merge conflict. It has now been merged.

PR #1223, opened 2021-10-26
allow unknown number of spaces in modprobe blacklists

### LINKED ISSUE #1219: Determine available network protocols - false negative results
**Describe the bug**
Lynis returns the suggestion:
"Determine if protocol 'rds' is really needed on this system [NETW-3200]"

There is an entry in the file:
`install rds    /bin/true`

The test returns a false negative because there are several "redundant" spaces between the word rds and /bin/true.

Correct command to search for entries:
`grep --files-with-matches --no-messages "^install \+rds \+/bin/true" /etc/modprobe.d/*`

**Version**
 - Distribution Ubuntu 20.04
 - Lynis version 3.0.6

**Expected behavior**
Search for entries with any number of spaces between words.

**Output**

**Additional context**

konstruktoid:
This PR:
- Allows unknown number of spaces when blacklisting kernel modules
- Closes #1219 

### DIFF
--- a/include/tests_filesystems
+++ b/include/tests_filesystems
@@ -619,7 +619,6 @@
                         Display --indent 2 --text "- Mount options of ${FILESYSTEM}" --result "${STATUS_PARTIALLY_HARDENED}" --color YELLOW
                         AddHP 4 5
                     else
-                        # if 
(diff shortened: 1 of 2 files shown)


mboelen:
Thanks, merged!

PR #1225, opened 2021-11-08
add symlink support to HasCorrectFilePermissions and IsWorldWritable

### LINKED ISSUE #1214: Some tests incorrectly recognise access rights
**Describe the bug**
Some tests incorrectly identify access rights when the target object is a symlink.

The problem is function: IsWorldWritable

**Version**
 - Distribution Ubuntu 20.04
 - Lynis version 3.0.6

**Expected behavior**
Correct recognition of the write right when the target object is a symlink.

**Output**

**Additional context**
We can test the problem in this way:
```
mv /etc/crontab /mnt/crontab
ln -s /mnt/crontab /etc/crontab
lynis audit system 2>&1 | grep SCHD-7704
```

### LINKED ISSUE #1245: Improve symlink permission checks
**Bug / behaviour / output**

For symlinks, the Lynis check `FILE-7524` should not test the permissions on the symlink itself, but the permissions of the target, because only the target's permission ultimately matters for access.

```
# Test: FILE-7524
2020-06-14 10:01:11 Test: checking file/directory /etc/motd
2020-06-14 10:01:11 Test: checking if file /etc/motd has the permissions set to 644 or more restrictive
2020-06-14 10:01:11 Outcome: permissions of file /etc/motd are not matching expected value (777 != rw-r--r--)

# why 777? because it's a symlink
$ ls -la /etc/motd
lrwxrwxrwx. 1 root root 25 Dec 17 23:21 /etc/motd -> /var/lib/update-motd/motd

# but it's actually 644!
$ ls -la /var/lib/update-motd/motd
-rw-r--r--. 1 root root 121 Jun 14 09:02 /var/lib/update-motd/motd
```

Mind though that this could be different for BSD-like systems. I think on those systems you can put actual permissions on symlinks (differing from the target) -- not sure though!

**Expected behavior**
Make the target's permissions count and not the symlink permissions.

konstruktoid:
This PR:
- adds symlink support to HasCorrectFilePermissions and IsWorldWritable
- closes #1214 
- closes #1245 

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -1306,6 +1306,11 @@
         if [ $# -ne 2 ]; then Fatal "Incorrect usage of HasCorrectFilePermissions"; fi
         CHECKFILE="$1"
         CHECKPERMISSION_FULL="$2"
+        # Check for symlink
+        if [ -L ${CHECKFILE} ]; then
+            ShowSymlinkPath ${CHECKFILE}
+            if [ ! "${SYMLINK}" = "" ]; then CHECKFILE="${SYMLINK}"; fi
+        fi
         if [ ! -d ${CHECKFILE} -a ! -f ${CHECKFILE} ]; then
             return 2
         else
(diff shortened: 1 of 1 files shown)


incitem (approved):
This PR fixes https://github.com/CISOfy/lynis/issues/1245 for me.

mboelen:
Thanks, merged!

PR #1227, opened 2021-11-08
Update ru

zhelemysh:


mboelen:
Thanks for the translation of the strings, much appreciated!

PR #1250, opened 2022-01-01
Dont write over FIND variable. Fixes #1021

### LINKED ISSUE #1021: False positive in PRNT-2308 checking CUPS listening in loopback
**Describe the bug**
6eb204a introduced a bug by overwriting the value of `FIND` with a possibly blank value that is later checked for its original value that no longer exists.

**Version**
 - Distribution Ubuntu 19.10
 - Lynis version 3.0.0

**Expected behavior**
PRNT-2308 is passed when it's configured to listen on TCP localhost only.

**Output**
```
Performing test ID PRNT-2308 (Check CUPSd network configuration)
Test: Checking CUPS daemon listening network addresses
Result: found network address: localhost:631
Result: CUPS daemon running on one or more interfaces (not limited to localhost)
```

**Additional context**
The bug is in this portion of code modified by 6eb204a that originally looked like this:
```bash
        FIND=$(${EGREPBINARY} "^(SSL)?Listen" ${CUPSD_CONFIG_FILE} | ${GREPBINARY} -v "/" | ${AWKBINARY} '{ print $2 }')
        COUNT=0
        for ITEM in ${FIND}; do
            LogText "Found network address: ${ITEM}"
            COUNT=$((COUNT + 1))
            FOUND=1
        done

        # Check if daemon is only running on localhost
        if [ ${FOUND} -eq 0 ]; then
            LogText "Result: no listen statement found in CUPS configuration file"
        elif [ ${COUNT} -eq 1 ]; then
            if [ "${FIND}" = "localhost:631" -o "${FIND}" = "127.0.0.1:631" ]; then
                LogText "Result: CUPS daemon only running on localhost"
                AddHP 2 2
            else
                LogText "Result: CUPS daemon running on one or more interfaces (not limited to localhost)"
                ReportSuggestion "${TEST_NO}" "Check CUPS configuration if it really needs to listen on the network"
                AddHP 1 2
            fi
```
`FIND` is populated with all the occurrences of `Listen` in `cupsd.conf`. If only one occurrence is found, it checks if that occurrence is `localhost:631` or `127.0.0.1:631` to pass the test.

6eb204a modified this to add a new portion of code after the first set of `FIND` but before checking its value that sets `FIND` to any occurrence of `Port` in `cupsd.conf`:
```bash
        # Search for Port statement
        FIND=$(${EGREPBINARY} "^Port 631" ${CUPSD_CONFIG_FILE})
        if [ -n "${FIND}" ]; then
            LogText "Result: found CUPS listening on port 631 (most likely all interfaces)"
            FOUND=1
        fi
```
If this new check doesn't return any value, `FIND` is set to an empty string and the task continues as before to check if only one `Listen` occurrence is stored in `FIND` and it points to `localhost`, but `FIND` doesn't store that value anymore but just an empty string, so this check will always fail.

JensChrG:
When searching for the"Port" statement, the result would overwrite the result from the "Listen" search.
This would lead to a false-positive result, when there was only a Listen statement for localhost.

Flipping the order of the two searches and keeping track of the first result in a separate variable fixes the problem.

### DIFF
--- a/include/tests_printers_spoolers
+++ b/include/tests_printers_spoolers
@@ -139,8 +139,18 @@
     Register --test-no PRNT-2308 --preqs-met ${PREQS_MET} --weight L --network NO --category security --description "Check CUPSd network configuration"
     if [ ${SKIPTEST} -eq 0 ]; then
         FOUND=0
-        # Checking network addresses
+        PORT_FOUND=0
+
         LogText "Test: Checking CUPS daemon listening network addresses"
+
+        # Search for Port statement
+        FIND=$(${EGREPBINARY} "^Port 631" ${CUPSD_CONFIG_FILE})
+        if [ -n "${FIND}" ]; then
+            LogText "Result: found CUPS listening on port 631 (most likely all interfaces)"
+            PORT_FOUND=1
+        fi
+
+        # Checking network addresses
         FIND=$(${EGREPBINARY} "^(SSL)?Listen" ${CUPSD_CONFIG_FILE} | ${GREPBINARY} -v "/" | ${AWKBINARY} '{ print $2 }')
         COUNT=0
         for ITEM in ${FIND}; do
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for your contribution, merged!

PR #1254, opened 2022-01-12
Check for pam_faillock in AUTH-9408

pyllyukko:
`pam_tally2` was removed in [PAM 1.5.0](https://github.com/linux-pam/linux-pam/releases/tag/v1.5.0) and `pam_faillock` has replaced it.

Also the check for `${ROOTDIR}etc/pam.conf` is not sufficient to determine if PAM is being used (at least in upcoming Slackware 15.0) so I added additional check for `${ROOTDIR}etc/pam.d` directory.

### DIFF
--- a/include/tests_authentication
+++ b/include/tests_authentication
@@ -1533,31 +1533,49 @@
     # Description : Logging of failed login attempts
     Register --test-no AUTH-9408 --weight L --network NO --category security --description "Logging of failed login attempts"
     if [ ${SKIPTEST} -eq 0 ]; then
-        if [ -f "${ROOTDIR}etc/pam.conf" ]; then
+        if [ -f "${ROOTDIR}etc/pam.conf" -o -d "${ROOTDIR}etc/pam.d" ]; then
             FOUND_PAM_TALLY2=0
             FOUND_TALLYLOG=0
-            if [ -s "${ROOTDIR}var/log/tallylog" ]; then
+            FOUND_PAM_FAILLOCK=0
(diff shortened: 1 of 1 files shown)


konstruktoid:
+1

Note that on Ubuntu `/etc/security/faillock.conf` is used for configuration and should be checked for any settings.
Example https://github.com/konstruktoid/hardening/blob/master/scripts/password#L21-L31

pyllyukko:
Indeed and the same configuration is also in Debian, Slackware and CentOS. But `pam_faillock` also has somewhat sane defaults, so checking that file would be required only if Lynis would require some explicit values for the settings.

borislavba:
That's great! 
Could somebody please just change
`if [ -f ${ROOTDIR}etc/login.defs ]; then`
to 
`if [ AUTH_FAILED_LOGINS_LOGGED -eq 1 -a -f ${ROOTDIR}etc/login.defs ]; then`
otherwise there are maximum points, but the result is that logging is disabled if there is no FAILLOG_ENAB yes in /etc/login.defs (I guess if pam_faillock is set up there no need for the line in login.defs ?).
P.s. Sorry if there are errors in the syntax, but you'll get the idea :)

pyllyukko:
> otherwise there are maximum points, but the result is that logging is disabled if there is no FAILLOG_ENAB yes in /etc/login.defs (I guess if pam_faillock is set up there no need for the line in login.defs ?).

`FAILLOG_ENAB` is not needed and not recognized at all by PAM.

PR #1260, opened 2022-01-23
perf_event_paranoid -> 2|3

pyllyukko:
The value of `3` was introduced in c76a4ca and the source was this[1].
The documentation in the source notes that the value of `3` requires a
patch for the Linux kernel. Vanilla kernel has the "maximum" value of
`2`[2].

[1] https://docs.clip-os.org/clipos/kernel.html#sysctl-security-tuning
[2] https://www.kernel.org/doc/html/v5.7/admin-guide/sysctl/kernel.html#perf-event-paranoid


mboelen:
Thanks, merged!

PR #1263, opened 2022-01-25
KRNL-5830: Check for /var/run/needs_restarting

pyllyukko:
Creating this file was introduced in [slackpkg](https://slackpkg.org/) version 15.0.10.


mboelen:
Thanks, merged!

Issue #1264, opened 2022-02-04
KRNL-5820: Add support for openSUSE's alternative security/limits.conf path


rfrohl:
**Is your feature request related to a problem? Please describe.**
openSUSE distros (MicroOS, Tumbleweed) provide an alternative way to update the system, called atomic updates. Part of this feature are alternative paths for config files under /usr/etc.

**Describe the solution you'd like**
KRNL-5730 should be able to detect limits.conf in none default locations, like /usr/etc/security/limits.conf.

**Required changes**
KRNL-5730 needs to check under /usr/etc/security/ for a limits.conf too.

**Additional context**
Details about the move to /usr/etc

https://kubic.opensuse.org/blog/2019-12-05-usr-etc/

rfrohl:
Posted a suggested change, please let me know if it would require changes.

PR #1265, opened 2022-02-04
[KRNL-5820] additional path for security/limits.conf

rfrohl:
Changes the user facing output to display a full path, allowing the user to
better grasp which security/limits.conf file is affected.

fix issue #1264


mboelen:
Thanks, it has been merged. Will apply a few changes so that ROOTDIR is not used within the test itself.

PR #1292, opened 2022-05-04
Only test Compression if sshd version < 7.4

### LINKED ISSUE #1291: misleading suggestion/false positive SSH-7408:Compression
**Describe the bug**

https://github.com/CISOfy/lynis/blob/master/include/tests_ssh#L138 suggests setting `Compression no` in `/etc/ssh/sshd_config` to prevent possible pre-authentication attacks on connections with compression enabled.

https://github.com/CISOfy/lynis/issues/217#issuecomment-358278113:

> Support for pre-authentication compression has been removed on on Sep 28, 2016 by the commit
https://github.com/openssh/openssh-portable/commit/0082fba4efdd492f765ed4c53f0d0fbd3bdbdf7f#diff-267d507f9cf4a70e051aaeecb89ad93b

> Since delayed is a legacy synonym for yes.

So any of `yes`, `no`, `delayed` can be considered secure on openssh installations that have been updated in the past 4 years.

**Version**
- false positive on [Debian stable](https://manpages.debian.org/bullseye/openssh-server/sshd_config.5.en.html#Compression)
- false positive on [Debian oldstable](https://manpages.debian.org/buster/openssh-server/sshd_config.5.en.html#Compression)
- false positive on [Debian oldoldstable](https://manpages.debian.org/stretch/openssh-server/sshd_config.5.en.html#Compression)
- Lynis version `master`

**Expected behavior**

The check for this value should be removed https://github.com/CISOfy/lynis/blob/master/include/tests_ssh#L138

**Output**

```
suggestion[]=SSH-7408|Consider hardening SSH configuration|Compression (set YES to NO)
```

**Additional context**

This check could be made more accurate by checking the actual `sshd` version, and comparing it with known versions with the patch applied... but I think this is unnecessarily complex; the question boils down to whether very old openssh versions should be supported by lynis.

konstruktoid:
This PR:
- Removes the Compression test for `sshd` versions > 7.4
- Closes #1291

### DIFF
--- a/include/tests_ssh
+++ b/include/tests_ssh
@@ -135,7 +135,6 @@
         SSHOPS="AllowTcpForwarding:NO,LOCAL,YES:=\
                 ClientAliveCountMax:2,4,16:<\
                 ClientAliveInterval:300,600,900:<\
-                Compression:NO,,YES:=\
                 FingerprintHash:SHA256,MD5,:=\
                 GatewayPorts:NO,,YES:=\
                 IgnoreRhosts:YES,,NO:=\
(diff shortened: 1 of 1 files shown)


mboelen:
Merged, thanks!

PR #1296, opened 2022-05-18
use systemctl get-default instead of following link

### LINKED ISSUE #1286: Systemd runlevel check not working in debian, ubuntu and archlinux
https://github.com/CISOfy/lynis/blob/64758216f20bfb198ae45a258fb2bb9e8ecdbc45/include/tests_kernel#L42-L44

The above check doesn't work in:
- Debian 8
- Debian 11
- Ubuntu 20.04
- Archlinux
- Manjaro

Due to that file not being present in them in a default install.

It does work in recent RHEL and Fedora versions.

Using `systemctl get-default` would be a perfect replacement because if the system uses systemd it must be present unlike the text file.

konstruktoid:
This PR:
- uses `${SYSTEMCTLBINARY} get-default`, as suggested by @marderbot, instead of following a link
- removes end of line spaces
- closes #1286 


mboelen:
Makes sense, so merged!

PR #1343, opened 2022-10-28
Replace newer tr command syntax with older ascii specific operations

alangonzalez:
For busybox support, some commands don't support the newer syntax :alnum: :cntrl: etc options.

I replaced them with the older counterparts and in one case replaced upper->lower lower->upper with `awk` replacements. 

Only issue I can see if we are allowing for LOCALE specific optimizations for using `tr` with the posix callbacks. I suspect we aren't since in the `grep` statement on line 71 we use the ASCII syntax of a-z.

So I believe this patch will preserve functionality for all platforms, but if there is concerns on still using :alnum: I can submit revised patch where I check if the `tr` command supports advanced functionality and then everytime the `tr` command is about to run put an if statement. I thought this would be klunky given how many uses of `tr`

The test for `tr` functionality is like this:
```
DATA=$(echo testing | tr -d '[:alnum:]')
if [ -n "${DATA}" ]; then
 # tr doesn't work
fi

### DIFF
--- a/include/profiles
+++ b/include/profiles
@@ -56,7 +56,7 @@
         fi
 
         # Security check for unexpected and possibly harmful escape characters (hyphen should be listed as first or last character)
-        DATA=$(grep -Ev '^$|^ |^#|^config:' "${PROFILE}" | tr -d '[:alnum:]/\[\]\(\)_\|,\.:;= \n\r-')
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for your work on this. Looks good to me, so I have merged it.

PR #1345, opened 2022-11-17
Added Slackware support for PackageIsInstalled() function

pyllyukko:
Slackware's `pkgtools` doesn't seem to have "is this package installed" functionality, but all the packages can be seen in `/var/lib/pkgtools/packages/` (though that path was introduced/changed in the recent [Slackware 15.0](https://ftp.osuosl.org/pub/slackware/slackware64-15.0/ChangeLog.txt)).

E.g. now with the `INSE-8000` test it now sees the file `inetd-1.79s-x86_64-14` (the suffix is version-arch-revision) and flags it as installed.


mboelen:
Thanks, merged! Sorry it took so long...

PR #1351, opened 2022-12-08
Change xargs -i to xargs -I

strangelittlemonkey:
From the man page:

-i[replace-str], --replace[=replace-str]
        This option is a synonym for -Ireplace-str if replace-str
        is specified.  If the replace-str argument is missing, the
        effect is the same as -I{}.  This option is deprecated;
        use -I instead.

The -i flag is not present on BSD xargs and is deprecated in GNU xargs.

### DIFF
--- a/include/tests_databases
+++ b/include/tests_databases
@@ -85,7 +85,7 @@
         LogText "Test: Trying to login to local MySQL server without password"
 
         # "-u root --password=" avoids ~/.my.cnf authentication settings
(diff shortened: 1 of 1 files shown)


gwollman:
This cannot work: the option-argument to `xargs -I` is mandatory. (That is why POSIX.2 introduced `-I` in the first place: the Utility Syntax Guidelines forbid optional option-arguments.)

mboelen:
Not merging this one, due to the note above and listed in https://github.com/CISOfy/lynis/pull/1396.

Thanks both for your guidance!

PR #1364, opened 2023-01-03
KRNL-5820: Accept ulimit with -H parameter also

pyllyukko:
```
ulimit: ulimit [-SHabcdefiklmnpqrstuvxPT] [limit]
```


mboelen:
Trying to understand this one. Wouldn't this alter the check or mix things up between multiple parameters?

pyllyukko:
> Trying to understand this one. Wouldn't this alter the check or mix things up between multiple parameters?

So the modification would allow (explicit) `ulimit -Hc` ("use the \`hard' resource limit") instead of only `ulimit -c`.

mboelen:
Thanks, merged.

PR #1366, opened 2023-01-07
Fix EOL dates.

noci2012:
CentOS 8 is EOL,  7 is not EOL...
Debian updated
Mageia supplemented (9 is not available yet)
OpenBSD supplemented


mboelen:
Thanks, merged!

PR #1373, opened 2023-02-23
tests_mac_frameworks: check for existance of semanage binary

jsegitz:
It's not installed everywhere, causing this error message:
/usr/share/lynis/include/tests_mac_frameworks: line 161: permissive: command not found
/usr/share/lynis/include/tests_mac_frameworks: line 162: permissive: command not found


mboelen:
Thanks! Made a minor change, to do the test if the binary value is not empty.

Apologies for the delay in merging.

Issue #1338, opened 2022-09-27
egrep: warning: egrep is obsolescent; using grep -E


Obema12:
The execution of `lynis audit system` is accompanied by a whole bunch of messages related to grep
```
egrep: warning: egrep is obsolescent; using grep -E
  - Total without nodev:8 noexec:13 nosuid:6 ro or noexec (W^X): 13 of total 27
  - Disable kernel support of some filesystems
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
egrep: warning: egrep is obsolescent; using grep -E
```

Fix it at last.

mboelen:
On what operating system and version are you running Lynis?

Obema12:
Archlinux.

konstruktoid:
man grep(1) on Ubuntu 20.04.5:
"the variant programs egrep, fgrep and rgrep are the same as grep -E, grep -F, and grep -r, respectively.  These variants are deprecated, but are provided for backward compatibility."

mboelen:
[...]

PR #1382, opened 2023-04-08
WIP: replace `egrep` with `grep -E` on modern Linux distros

XenGi:
Ohai,
on Arch Linux, lynis starts to spit out these messages. They seem to come from a deprecation in grep. This is the warning:

```
egrep: warning: egrep is obsolescent; using grep -E
```

### DIFF
--- a/lynis
+++ b/lynis
@@ -217,10 +217,10 @@
 
     # Extract the short notation of the language (first two characters).
     if [ -x "$(command -v locale 2> /dev/null)" ]; then
-        LANGUAGE=$(locale | grep -R "^LANG=" | cut -d= -f2 | cut -d_ -f1 | tr -d '"' | grep -R "^[a-z]{2}$")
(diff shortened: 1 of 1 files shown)


konstruktoid:
Related https://github.com/CISOfy/lynis/issues/1338

Any plans to replace the `EGREPBINARY` as well?

konstruktoid:
There's an `EGREP` variable as well. See https://github.com/CISOfy/lynis/issues/1338#issuecomment-1399217324

konstruktoid:
Ah, but there is already an `EGREP` variable at https://github.com/CISOfy/lynis/blob/master/include/binaries#L172

XenGi:
Oh, that makes it a bit more complex. I only had a look into the main script. The problem is that even on Arch and other newer Linux distos it will find the egrep binary but then spit out these nasty warnings.

XenGi:
I'll probably won't find time to work on this.

Issue #1322, opened 2022-08-14
Malware scanner MDATP not recognized


Vilican:
**Describe the bug**
Microsoft Defender ATP is not recognized (although previously fixed under issue #992 and PR #994)

**Version**
 - Debian Bullseye
 - Lynis version 3.0.8

**Output**
```
2022-08-14 22:57:56 Starting Lynis 3.0.8 with PID 624075, build date 2022-05-17
2022-08-14 22:57:56 ====
2022-08-14 22:57:56 ### 2007-2021, CISOfy - https://cisofy.com/lynis/ ###
2022-08-14 22:57:56 Checking permissions of /usr/share/lynis/include/profiles
2022-08-14 22:57:56 File permissions are OK
2022-08-14 22:57:56 Reading profile/configuration /etc/lynis/default.prf
2022-08-14 22:57:56 Action: created temporary file /tmp/lynis.ShJul38atZ
2022-08-14 22:57:56 Language set via profile to ''
2022-08-14 22:57:56 Plugin 'authentication' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'compliance' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'configuration' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'control-panels' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'crypto' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'dns' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'docker' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'file-integrity' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'file-systems' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'firewalls' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'forensics' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'hardware' enabled according profile (/etc/lynis/default.prf)
2022-08-14 22:57:56 Plugin 'intrusion-detection' enabled according profile (/etc/lynis/default.prf)
[...]

PR #1384, opened 2023-04-18
Adding test MALW-3291 in order to solve issue 1322 

xnoguer:
I added a new test (id MALW-3291) in order to solve issue 1322. This is related to issue 994.

A few notes:
- I could have just considered whether MDATPBINARY was set or not (in include/binaries) and use that as a condition to set (or unset) MALWARE_SCANNER_INSTALLED in include/tests_malware. But I thought that adding a new test for mdatp was a cleaner solution. If I was wrong, please don't hesitate to tell me so.
- Tried to respect what I think is the criteria for assigning ids to new tests. Again, if assigned a wrong id to the new test, please don't hesitate to tell me so.
- Tested as thoroughly as I could (Debian 11), but more testing would be welcome.


alexander-krug:
@xnoguer I am having the same issue as you and #1322 and like your clean approach to using the "malware_scanner_installed" variable to solve the problem. 

I performed a recent run on my system with mdatp installed, enabled and running. Lynis reports "Result: no malware scanner found" just as #1322. 

 /var/log/lynis.log shows: 
`2023-09-02 20:38:29 Starting Lynis 3.0.6 with PID 2823, build date 2021-07-22
(...)
2023-09-02 20:38:36   Found known binary: mdatp (Microsoft Defender ATP, malware scanner) - /usr/bin/mdatp
(...)
2023-09-02 20:39:21 Performing test ID HRDN-7230 (Check for malware scanner)
2023-09-02 20:39:21 Test: Check if a malware scanner is installed
2023-09-02 20:39:21 Result: no malware scanner found
`

/var/log/lynis-report.dat shows: 
`malware_scanner_installed=0`

https://github.com/CISOfy/lynis/blob/master/include/tests_malware is missing a test to set malware_scanner_installed to 1 if mdatp is installed. This commit should fix this.

GuizmoPhil:
I'm having the same issue with Lynis version 3.1.1. MDATP (Microsoft Defender) isn't recognized as a malware scanner.

System is running Ubuntu 20.04, which is based on Debian 10.

mboelen:
Looks good, so I have merged it.

PR #1385, opened 2023-04-20
Fixing "Security check failed" message for unprivileged mode. Issue 1383

xnoguer:
Making sure that the "Security check failed" message is also shown when running in non-privileged mode (not root). A few notes:

- I included data_upload among the 6 files to be checked, but I cannot test it.
- I added a check in the function SafeFile(), only for non-privileged mode. Didn't want to add a new function for the checks of this section of code (lines 147 to 214) because it seemed to disruptive to me, and would have created a chicken and egg problem in line 218 (". ${INCLUDEDIR}/functions").
- I tested this in Debian 11.

### DIFF
--- a/include/functions
+++ b/include/functions
@@ -3103,6 +3103,11 @@
                     unsafe=1
                     LogText "Security alert: file is not owned by active user, but can write to it"
                 fi
+                # File is not owned by active user, and not readable by him
+                if [ ! -O "${FILE}" -a ! -r "${FILE}" ]; then
+                    unsafe=1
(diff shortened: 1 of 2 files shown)


mboelen:
Can't merge this one, as it performs a double task. The consts/functions files are the only ones that need to be checked from within the main program, as the other inclusions are checked with the related function. So we have to come up with an alternative way.

As the file permissions have been adjusted in the tarball, I suggest to make no change at this moment. For that reason I will close it, also as part of a cleanup session.

PR #1387, opened 2023-04-25
Checking for errors, not only warning in docker info output

xnoguer:
When running lynis audit system as a non-privileged user, and assuming dockerd running as root, lynis will report an ERROR por Docker status but will report NONE for "Docker info output (warnings)", not reporting the error ("Docker daemon socket") that docker info reported. 

This pull request makes sure that error is reported and the hardening points are consistent with that error.


mboelen:
Thanks, looks good. Merged. Sorry it took so long...

Issue #1367, opened 2023-01-08
[FILE-6344] Invalid check on the kernel version


pyllyukko:
**Describe the bug**
There's a bug in the check for prerequirements where the kernel version is checked:

```bash
        LINUX_KERNEL_MAJOR=$(echo $OS_KERNELVERSION | ${AWKBINARY} -F. '{print $1}')
        LINUX_KERNEL_MINOR=$(echo $OS_KERNELVERSION | ${AWKBINARY} -F. '{print $2}')
        if [ -n "${LINUX_KERNEL_MAJOR}" -a -n "${LINUX_KERNEL_MINOR}" ]; then
            if [ ${LINUX_KERNEL_MAJOR} -ge 3 -a ${LINUX_KERNEL_MINOR} -ge 3 ]; then PREQS_MET="YES"; else PREQS_MET="NO"; fi
```

It should check that the kernel is newer than 3.3, but the test is written as such, that all minor versions below 3 will fail this test, even though the major version would be > 3. E.g. 4.[012], 5.[012] etc.

**Version**
 - Lynis version 3.0.8

**Expected behavior**
E.g. Linux kernel version 6.0.0 would fulfil the prerequirements.

**Output**
Prerequirements are not met if kernel's minor version is below 3.

xnoguer:
@pyllyukko Could you please try this pull request?
https://github.com/CISOfy/lynis/pull/1388

pyllyukko:
> @pyllyukko Could you please try this pull request?

Yes, it fixes this issue. Thanks.

PR #1388, opened 2023-04-26
Fixing test on kernel major and minor version (FILE-6344). Issue 1367

xnoguer:
This pull request fixes issue 1367: https://github.com/CISOfy/lynis/issues/1367

Tested on:
- Debian 11, kernel 5.10.0. Failure and success.
- Ubuntu 19.04, kernel 5.0, which skipped the test because on minor version 0. Failure and success.

### DIFF
--- a/include/tests_filesystems
+++ b/include/tests_filesystems
@@ -346,7 +346,13 @@
         LINUX_KERNEL_MAJOR=$(echo $OS_KERNELVERSION | ${AWKBINARY} -F. '{print $1}')
         LINUX_KERNEL_MINOR=$(echo $OS_KERNELVERSION | ${AWKBINARY} -F. '{print $2}')
         if [ -n "${LINUX_KERNEL_MAJOR}" -a -n "${LINUX_KERNEL_MINOR}" ]; then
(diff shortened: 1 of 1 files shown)


Smankusors (approved):
This is happened on my system (kernel 6.1), and I verified that this PR changes fixes the original issue.

wdyt @mboelen ?

mboelen:
Thanks, merged!

@xnoguer should we also add a "catch" for unspecified future versions? For example, showing an exception if the kernel is at higher levels, or not needed as support for the checked options is now covered fully, including future-proof testing?

xnoguer:
@mboelen I don't think it's needed. If the test didn't make sense from a future kernel release onward (let's say X.Y), then we would have to change the code so that PREQS_MET for test FILE-6344 is only set to "YES" when kernel version is between 3.3 and X.Y.
But there's no way to know this now (as far as I'm aware). In fact it would mean that the hidepid option stops being used or adds new possible values force us to rewrite test FILE-6344.

mboelen:
Yup, makes sense. Was also my thought, but wanted to double-check with you.

Thanks :+1:

Issue #1354, opened 2022-12-17
Ubuntu - pam_cracklib; pam_passwdqc no longer available


vk6xebec:
**Describe the bug**
pam_cracklib; pam_passwdqc and apt-listbugs are no longer available on Ubuntu. 

**Version**
 - Distribution: Ubuntu 22.10
 - Lynis version: 3.0.7+

**Expected behavior**
The check does not throw up an suggestion or error

**Output**
Please disable this check on Ubuntu, or maybe suggest an alternative if there is one

**Context**
  * Install a PAM module for password strength testing like pam_cracklib or pam_passwdqc [AUTH-9262] 
      https://cisofy.com/lynis/controls/AUTH-9262/

Note that this is also applicable for
  * Install apt-listbugs to display a list of critical bugs prior to each APT installation. [DEB-0810] 
      https://cisofy.com/lynis/controls/DEB-0810/

vk6xebec:
Found the package. Please change the script to refer to the following package instead: libpam-passwdqc

xnoguer:
@vk6xebec Can you try this pull request ?
https://github.com/CISOfy/lynis/pull/1389

mboelen:
Merged PR by @xnoguer , will be part of next release.

PR #1389, opened 2023-04-27
Change in suggestion message. Issue 1354

xnoguer:
This pull request just adds the name of the latest package (libpam-passwdqc) for checking password strength used in Debian and Ubuntu distributions. Solves issue https://github.com/CISOfy/lynis/issues/1354

Run test AUTH-9262 with and without the package just to make sure the behaviour was as expected. Debian 12.


mboelen:
Thanks, merged! Sorry it took so long to merge.

PR #1390, opened 2023-05-03
osdetection: add OpenEmbedded and Poky

akuster:
This allows detection the OpenEmbedded "nodistro" and The Yocto Projects distro "Poky"


akuster:
If this is going to be rejected, please let me know.

mboelen:
Thanks, merged!

PR #1395, opened 2023-05-18
Adding missing test number for suggestion. Issue 1280

### LINKED ISSUE #1280: TEST_NO missing for a suggestion in tests_nameservices
**Describe the bug**
All ReportSuggestion lines have ${TEST_NO} as first argument, except one in tests_nameservices file

The following command shoud return nothing, but return 1 line :
grep -r ReportSuggestion | grep tests | grep -v TEST_NO
tests_nameservices:                ReportSuggestion "Disable the usage of NIS/NIS+ and use an alternative like LDAP or Kerberos instead"

Please change line to:
 ReportSuggestion "${TEST_NO}" "Disable the usage of NIS/NIS+ and use an alternative like LDAP or Kerberos instead"

Regards,
SixK

xnoguer:
- This PR closes #1280 
- Spent sometime trying to get ypbind up and running but couldn't. So this is not tested.

### DIFF
--- a/include/tests_nameservices
+++ b/include/tests_nameservices
@@ -485,7 +485,7 @@
                 LogText "Result: ypldap is running"
                 Display --indent 2 --text "- Checking ypldap status" --result "${STATUS_FOUND}" --color GREEN
             else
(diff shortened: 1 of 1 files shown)


mboelen:
Good catch! Merged.

PR #1396, opened 2023-05-19
Replacing deprecated option -i. Issue 1300

### LINKED ISSUE #1300: xargs -i is nonstandard
`include/tests_databases` invokes `xargs` with the nonstandard `-i`(lower-case `i`) flag, resulting in an error message to stderr. It should use the standard `-I` (upper-case `I`) flag, which requires an argument instead (but might need logic to deal with nonstandard `xargs` implementations). The GNU `xargs` manual page claims that `-I{}` does what the nonstandard `-i` flag did.  I can confirm that this eliminates the error message, but not that it is doing the expected thing.

**Version**
- 8d9cdb22f413f2b2c941ae4750ca40411ce8a194 (tagged 3.0.8)
- FreeBSD 12.3

**Expected behavior**
`xargs` should be invoked with valid command-line arguments so it does not write a diagnostic and exit with an error status.

**Output**
```
xargs: invalid option -- i
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements] [-S replsize]]
             [-J replstr] [-L number] [-n number [-x]] [-P maxprocs]
             [-s size] [utility [argument ...]]
```

**Additional context**

xnoguer:
- I assume the issue reported ( #1300 ) is related to this (from xargs manual page):
       _-i[replace-str], --replace[=replace-str]
              This option is a synonym for  -Ireplace-str  if  replace-str  is
              specified.   If  the replace-str argument is missing, the effect
              is the same as -I{}.  This option is deprecated; use -I instead_
- Tested in Debian. Can't test on Freebsd.
- This PR closes #1300

### DIFF
--- a/include/tests_databases
+++ b/include/tests_databases
@@ -216,8 +216,8 @@
 
     Register --test-no DBS-1828 --preqs-met ${PREQS_MET} --weight L --network NO --category security --description "Test PostgreSQL configuration"
     if [ ${SKIPTEST} -eq 0 ]; then
-        FIND_PATHS="${ROOTDIR}etc/postgres ${ROOTDIR}var/lib/postgres/data ${ROOTDIR}usr/local/pgsql/data"
-        CONFIG_FILES=$(${FINDBINARY} -L ${FIND_PATHS} -type f -name "*.conf" -print0 2> /dev/null | ${TRBINARY} -cd '[:print:]\0' | ${TRBINARY} -d '\n' | ${TRBINARY} '\0' '\n' | xargs -i sh -c 'test -r "{}" && echo "{}"' | ${SEDBINARY} "s/ /:space:/g")
+        FIND_PATHS="${ROOTDIR}etc/postgres ${ROOTDIR}etc/postgresql ${ROOTDIR}var/lib/postgres/data ${ROOTDIR}usr/local/pgsql/data"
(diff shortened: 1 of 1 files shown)


gwollman:
Confirmed that this suppresses the error in FreeBSD.  However, since I submitted issue 1300, the FreeBSD packaging as switched to versioned `data` directories, so `FIND_PATHS` needs to be altered to match them in order for the test to actually find the configuration files it's looking for,

gwollman:
(For the avoidance of doubt: `data*` works, but `data[[:digit:]][[:digit:]]` would be safer, and for complete confidence you should probably check for a `postgresql_data` rc configuration variable using something like `$(sysrc -nq postgresql_data || echo /usr/local/pgsql/data[[:digit:]][[:digit:]])`.

xnoguer:
@gwollman FIND_PATHS already contains the path /usr/local/pgsql/data in the master branch and also in the last released version of Lynis (3.0.8).
- Are you using a version of Lynis older than 3.0.8, and if so what version?
- If test DBS-1828 still fails, can you post here the relevant lines of lynis.log so I can take a look ?

xnoguer:
@gwollman I just saw issue #1397, so now I understand the problem with the data path is clearly a different problem than the one from issue #1300. Ignore my previous questions. I'll try to submit a pull request for issue 1397.

strangelittlemonkey:
This duplicates https://github.com/CISOfy/lynis/pull/1351

xnoguer:
@strangelittlemonkey Agreed. Both pull requests are similar. Thanks.

gwollman:
This one is correct, #1351 is not (it just changes the flag character without supplying the option-argument).

mboelen:
Thanks, implemented and closed #1351.

PR #1404, opened 2023-05-30
Add malware detection support for SentinelOne

avenjamin:
Initial commit to detect SentinelOne as 'Malware scanner'
Need to verify detection on more OS's


mboelen:
Thanks, merged (with changes)!

Issue #1393, opened 2023-05-11
Unknown OS: NXP LSDK


p47-6:
The operating system from NXP "Layerscape SDK" is not recognized.

Here is the output from `cat /etc/os-release`:
```
NAME="NXP LSDK"
VERSION=2004
ID=lsdk
PRETTY_NAME="NXP LSDK 2004 main"
```
For reference, here is the link to their product: https://www.nxp.com/design/software/embedded-software/linux-software-and-development-tools/layerscape-linux-distribution-poc:LAYERSCAPE-SDK

PR #1409, opened 2023-06-10
Update osdetection with LSDK

digitalcheetah:
This should close issue #1393. It does need to be tested since I don't have access to LSDK.


mboelen:
Thanks, merged!

PR #1424, opened 2023-07-30
use grep -E instead of egrep if possible

xambroz:
Hello,
recent /usr/bin/egrep on Linux systems is throwing annoying warning that egrep is obsolete and that grep -E is used instead. 
To avoid this warning the EGREPBINARY is set to "${GREPBINARY} -E" if it has extended regexp in the "--help"
Best regards
Michal Ambroz

### DIFF
--- a/include/binaries
+++ b/include/binaries
@@ -337,6 +337,15 @@
         Report "binaries_sgid_count=${SGID_BINARIES}"
         Report "binary_paths=${BINARY_PATHS_FOUND}"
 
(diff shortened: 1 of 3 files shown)


konstruktoid:
https://github.com/CISOfy/lynis/issues/1338

mboelen:
I have this feeling this will break on something like Solaris. Most likely `--help` is not even supported by those older implementations. At the same time, we have to move forward, so I will merge this.

D3vil0p3r:
@mboelen The latest lynis stable version does not contain this warning fix, right?

Issue #1418, opened 2023-06-23
tests_firewalls: line 551: STATUS_NOT_ACTIVE: unbound variable


wileyhy:
**Describe the bug**
Similar bugs: #1416 , #1415 , #1414  , #1413  and #1333  .

**Version**
Distribution: Fedora 38

** # `uname -a`

>     Linux localhost-live 6.2.9-300.fc38.x86_64 https://github.com/CISOfy/lynis/issues/1 SMP PREEMPT_DYNAMIC Thu Mar 30 22:32:58 UTC 2023 x86_64 GNU/Linux

** # `lynis --version`

>     egrep: warning: egrep is obsolescent; using grep -E
>     egrep: warning: egrep is obsolescent; using grep -E
>     3.0.8
> 

**Expected behavior**
No warnings.

**Output**

> [DEBUG] Performing test ID FIRE-4524 (Check for CSF presence)
> [DEBUG] Performing test ID FIRE-4590 (Check firewall status)
> /usr/share/lynis/include/tests_firewalls: line 551: STATUS_NOT_ACTIVE: unbound variable
>

wileyhy:
[lynis-bug-1418-unified-diff.txt](https://github.com/CISOfy/lynis/files/11850720/lynis-bug-1418-unified-diff.txt)

mboelen:
Considering this already resolved with recent changes and PR https://github.com/CISOfy/lynis/pull/1426

It not, let me know.

mboelen:
Addition: Variable now set in include/consts with commit https://github.com/CISOfy/lynis/commit/1a26005b506d2c5dbf0ce2b403dd3b612b2cb374

Issue #1414, opened 2023-06-22
tests_insecure_services: line 165: STATUS_NOT_ACTIVE: unbound variable


wileyhy:
While trying to debug a shellcheck issue, SC2240 a number of these "unbound variable" bugs have been halting the program. 

** # `diff -u /usr/share/lynis/include/tests_insecure_services_orig /usr/share/lynis/include/tests_insecure_services`

> --- /usr/share/lynis/include/tests_insecure_services_orig	2022-05-16 17:00:00.000000000 -0700
> +++ /usr/share/lynis/include/tests_insecure_services	2023-06-22 08:36:47.753368860 -0700
> @@ -162,7 +162,7 @@
>              XINETD_ACTIVE=1
>          else
>              LogText "Result: xinetd is NOT running"
> \-            Display --indent 4 --text "- xinetd status" --result "${STATUS_NOT_ACTIVE}" --color GREEN
> \+            Display --indent 4 --text "- xinetd status" --result "${STATUS_NOT_ACTIVE:-"Warning: unbound variable STATUS_NOT_ACTIVE, tests_insecure_services, line $LINENO"}" --color GREEN
>          fi
>      fi
>  \#

mboelen:
Should be resolved in next release with recent changes (see linked #1426)

Issue #1413, opened 2023-06-22
tests_networking: line 695: STATUS_NOT_ACTIVE: unbound variable


wileyhy:
**Describe the bug**
It looks like the developer profile enables `set -u`, which is fine, of course, but it halts the program. I'm trying to debug a separate issue, regarding SC2240, but these other bugs are cropping up. 

**Version**
 - Fedora 38
 - lynis version 3.0.8

**Expected behavior**
I expected for there to not be any `set -u` bugs. 

**Output**

> /usr/share/lynis/include/tests_networking: line 695: STATUS_NOT_ACTIVE: unbound variable
>

wileyhy:
** # `diff -u /usr/share/lynis/include/tests_networking /usr/share/lynis/include/tests_networking_orig`

> --- /usr/share/lynis/include/tests_networking	2023-06-22 08:30:35.635552582 -0700
> +++ /usr/share/lynis/include/tests_networking_orig	2022-05-16 17:00:00.000000000 -0700
> @@ -692,7 +692,7 @@
>              Display --indent 2 --text "- Checking status DHCP client" --result "${STATUS_RUNNING}" --color WHITE
>              DHCP_CLIENT_RUNNING=1
>          else
> \-            Display --indent 2 --text "- Checking status DHCP client" --result "${STATUS_NOT_ACTIVE:-"Warning: unbound variable STATUS_NOT_ACTIVE, tests_networking, line $LINENO"}" --color WHITE
> \+            Display --indent 2 --text "- Checking status DHCP client" --result "${STATUS_NOT_ACTIVE}" --color WHITE
>          fi
>      fi
>  #
>

wileyhy:
** # `diff -u /usr/share/lynis/include/tests_networking_orig /usr/share/lynis/include/tests_networking`

> --- /usr/share/lynis/include/tests_networking_orig	2022-05-16 17:00:00.000000000 -0700
> +++ /usr/share/lynis/include/tests_networking	2023-06-22 08:30:35.635552582 -0700
> @@ -692,7 +692,7 @@
>              Display --indent 2 --text "- Checking status DHCP client" --result "${STATUS_RUNNING}" --color WHITE
>              DHCP_CLIENT_RUNNING=1
>          else
> \-            Display --indent 2 --text "- Checking status DHCP client" --result "${STATUS_NOT_ACTIVE}" --color WHITE
[...]

PR #1426, opened 2023-08-02
Fix STATUS_NOT_ACTIVE: unbound variable

amkorb:
Concerns issues n° #1413 #1414 #1418 

 During my audit, i encountered a white space at the end of "Checking status DHCP client" in tests_networking instead of the result message. With the strict mode raised, I got a "STATUS_NOT_ACTIVE: unbound variable" error as reported in the issues above. 

The error seems to arise from the absence of "STATUS_NOT_ACTIVE" variable declaration in the languages files. 
I've added "STATUS_NOT_ACTIVE="NOT ACTIVE"" to the 22 languages files to resolve this issue.


mboelen:
Thanks, merged!

PR #1429, opened 2023-08-04
software-eol.db: Add new Debian/Ubuntu releases and update EOL dates

samueloph:
software-eol.db: Add new Debian/Ubuntu releases and update EOL dates


mboelen:
Thanks, merged (with changes)

PR #1434, opened 2023-08-27
[PKGS-7388] Fix security repository test for new apt sources format

silentcreek:
In addition to the traditional one-line-per-source style, apt supports a newer sources list style with multi-line-stanzas (DEB822 style). This new style is e.g. used in Debian's latest docker images and expected to be the default in Ubuntu 23.10. The current test fails to recognize those sources.

An example of the new style would be: 
Types: deb
URIs: https://deb.debian.org/debian-security
Suites: bookworm-security
Components: main non-free-firmware
Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg

Change the regular expression to match the newer style as well.

### DIFF
--- a/include/tests_ports_packages
+++ b/include/tests_ports_packages
@@ -1043,7 +1043,7 @@
             fi
             if [ -d /etc/apt/sources.list.d ]; then
                 LogText "Searching for security.debian.org/security.ubuntu.com or security repositories in /etc/apt/sources.list.d directory"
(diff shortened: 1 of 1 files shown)


silentcreek:
Oh, I forgot to explain:
The change only applies to the regex for the files in /etc/apt/sources.list.d/ but not to the test of /etc/apt/sources.list itself. The reason is that the new style requires the files in this format to have the ending .sources while the old style is expected in files ending with .list
While we could use two different regular expressions depending on the file ending, I didn't want to make the test overly complicated.

PR #1449, opened 2023-12-07
Added Alpine Linux EOL dates

kolenichsj:
Added EOL dates for Alpine linux as listed on https://www.alpinelinux.org/releases/ . Values generated using date "+%s" --utc --date=YYYY-MM-DD


mboelen:
Thanks, merged!

PR #1451, opened 2023-12-14
AIX EOL updates

st739:
EOL dates for current and obsolete versions of AIX.


mboelen:
The pull request included the full file instead of the changes. So it looks like something did not go well, but I've merged it anyways.

PR #1456, opened 2024-02-14
Added initial version of a Kerberos plugin

pyllyukko:
As it says in the title, this is only an initial version. I wanted to open this PR to also open the discussion on the Kerberos hardening topic.

All the feedback is very welcome.

I'm planning on extending this and adding tests for at least:

* Checking `krb5.conf` (clients) and `kdc.conf` (KDC) for hardened settings
* ~~Checking for keys with weak encryption types~~ -> 075e3a6
* Checking various security related attributes on different principals
    * The principals first need to be placed into different categories such as: hosts, users, services & krb5 internal etc.


mboelen:
Looks like a good start. The only "thing" is that this would be better to be placed in a category. I guess this could be Authentication or if you feel like it will be a good amount of tests, then we could introduce a new one "Kerberos".

Rationale to move it: these tests do more than just collect information (main purpose of plugins), and may actually provide suggestions/warnings. For that reason it is better to make it part of the core, instead of a plugin.

pyllyukko:
> Looks like a good start. The only "thing" is that this would be better to be placed in a category. I guess this could be Authentication or if you feel like it will be a good amount of tests, then we could introduce a new one "Kerberos".

Ok. I'll do that then. Surely this is related to authentication, but maybe it would be good to place it into it's own Kerberos category. Or what do you think?

> Rationale to move it: these tests do more than just collect information (main purpose of plugins), and may actually provide suggestions/warnings. For that reason it is better to make it part of the core, instead of a plugin.

Makes sense.

mboelen:
I would say, let's create a new section :rocket:

So my suggestions for the steps:
- The file in this PR can be moved to include/tests_kerberos
- Include a section header Kerberos
- The tests named/numbered with KRB-1234
- My suggestion for the numbering would be to start at KRB-1000 and increase, let's say in steps of 10
- Then include it from the main program

If possible, it would be good for the layout to first detect if Kerberos is there at all. If not, then let's at least show one line so that the section does not remain empty on the screen. If there is no Kerberos at all, then all remaining tests can be silent if they are skipped. Keeps the screen as clean as possible, unless there is actually something useful to share.

mboelen:
Merged!

mboelen:
I have updated the language files so they are aware of the translation variable as well.

The section need a few minor adjustments, as it now shows output even though I have no Kerberos on my test system. Will see what I can do tomorrow to improve the output.

pyllyukko:
> The section need a few minor adjustments, as it now shows output even though I have no Kerberos on my test system. Will see what I can do tomorrow to improve the output.

#1499

PR #1458, opened 2024-03-11
Added CRYP-7932 to determine if the system has enabled macOS FileVault.

florian-sonnenschein-gcx:
This code will check if macOS FileVault is enabled.

### DIFF
--- a/include/tests_crypto
+++ b/include/tests_crypto
@@ -218,6 +218,27 @@
 #
 #################################################################################
 #
+    # Test        : CRYP-7932
+    # Description : Determine if system has enabled macOS FileVault encryption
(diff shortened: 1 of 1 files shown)


mboelen (changes requested):
Looks like a good start! A few suggestions to make it better:

- Move the "filevault_status" directory into the `case`, as nothing is done with the variable
- Do a check if fdesetup exists. Maybe it does exist on macOS now, but that might change in the future.
- Add some hardening points (AddHP). I would suggest 3 of 3 when enabled, 0 of 3, when not.

mboelen:
Merged, thanks!

PR #1460, opened 2024-03-11
Added PKGS-7305 - macOS Applications and CoreService (Name & Version)

florian-sonnenschein-gcx:
Query macOS /Applications and Apple CoreServices (Name & Version) and add them to the Report.


mboelen:
Thanks, merged!

PR #1482, opened 2024-04-07
Update tests_php to include PHP8.3 and 8.4

vk6xebec:
[Update tests_php to include PHP8.3 and 8.4]


mboelen:
Thanks, merged! Also cleaned up PHP 5, as it is deprecated for a while.

PR #1489, opened 2024-04-29
changed wording of the output AllowUsers to AllowGroups in test SSH-7440

katalyst666:
In the evaluation of AllowGroups, AllowUsers was in the output text.  I have corrected this


mboelen:
Good catch, merged!

PR #1494, opened 2024-05-09
postgresDB fixes for almalinux, rockylinux and freebsd.

arikkert:
Tested on almalinux8/9, rockylinux9 and freebsd13.
Support postmaster as possible servicename.
Support correct (default) location per OS.
use new xargs -I syntax, else it breaks on FreeBSD.


mboelen:
Thanks, merged!

PR #1498, opened 2024-05-15
Fixed a typo

pyllyukko:


mboelen:
Does not look like a typo to me. This test performs a test of the group ownership.

pyllyukko:
> Does not look like a typo to me. This test performs a test of the group ownership.

True. Changed the commit message to "Rephrasing". As the message talks about user, but means group ownership.

mboelen:
Ah, yes... Merged!

PR #1499, opened 2024-05-15
KRB-1010: Silence output when SKIPTEST != 0

pyllyukko:
Fixes https://github.com/CISOfy/lynis/pull/1456#issuecomment-2113390171


mboelen:
Merged!

PR #1515, opened 2024-06-25
Accept Kerberos translation in DE Language

just-maik:
Kerberos is Kerberos in german. And its a brand name wich stays the same anyway.


mboelen:
Merged, thanks!

Issue #1496, opened 2024-05-14
[INSE-8000] The inetd service is not detected


ferorge:
**Is your feature request related to a problem? Please describe.**
When lynis runs the test [INSE-8000] it does not detect the inetd service that is running.
Consequently tests [INSE-8002], [INSE-8004] and [INSE-8016] are skipped and test [INSE-8006] returns an improvement suggestion.
OS: ubuntu 24.04
lynis:3.1.2

> 2024-05-14 05:29:07 ====
> 2024-05-14 05:29:07 Performing test ID INSE-8000 (Installed inetd package)
> 2024-05-14 05:29:07 Test: Checking if inetd is installed
> 2024-05-14 05:29:07 Result: inetd is NOT installed
> 2024-05-14 05:29:07 ====
> 2024-05-14 05:29:08 Skipped test INSE-8002 (Check for enabled inet daemon)
> 2024-05-14 05:29:08 Reason to skip: Prerequisites not met (ie missing tool, other type of Linux distribution)
> 2024-05-14 05:29:08 ====
> 2024-05-14 05:29:08 Skipped test INSE-8004 (Presence of inetd configuration file)
> 2024-05-14 05:29:08 Reason to skip: Prerequisites not met (ie missing tool, other type of Linux distribution)
> 2024-05-14 05:29:08 ====
> 2024-05-14 05:29:08 Performing test ID INSE-8006 (Check configuration of inetd when disabled)
> 2024-05-14 05:29:08 Test: check if all services are disabled when inetd is disabled
> 2024-05-14 05:29:08 Result: found services in inetd, even though inetd is not running
> 2024-05-14 05:29:08 Suggestion: Although inetd is not running, make sure no services are enabled in /etc/inetd.conf, or remove inetd service [test:INSE-8006] [details:-] [solution:-]
> 2024-05-14 05:29:08 ====

```
# systemctl status inetd
● inetutils-inetd.service - GNU Network Utilities internet superserver
     Loaded: loaded (/usr/lib/systemd/system/inetutils-inetd.service; enabled; preset: enabled)
     Active: active (running) since Mon 2024-05-13 05:29:48 -03; 24h ago
       Docs: man:inetutils-inetd(8)
             https://www.gnu.org/software/inetutils/manual/
   Main PID: 10340 (inetutils-inetd)
      Tasks: 1 (limit: 4425)
     Memory: 416.0K (peak: 1.0M)
        CPU: 67ms
[...]

Issue #1497, opened 2024-05-14
[PKGS-7410] Does not detect the package containing the kernel.


ferorge:
**Is your feature request related to a problem? Please describe.**
When lynis runs the test [PKGS-7410] it does not detect the package containing the kernel.
hardware: odroid xu4 hardkernel
OS: ubuntu 24.04 armhf
lynis:3.1.2

> 2024-05-14 05:28:28 ====
> 2024-05-14 05:28:28 Performing test ID PKGS-7410 (Count installed kernel packages)
> 2024-05-14 05:28:28 Test: Checking how many kernel packages are installed
> 2024-05-14 05:28:28 Result: found no kernels from dpkg -l output, which is unexpected
> 2024-05-14 05:28:28 Exception: test has an exceptional event (PKGS-7410) with text Could not find any kernel packages via package manager
> 2024-05-14 05:28:33 ====

```
# dpkg -S /boot/zImage
linux-odroid-5422: /boot/zImage

```

**Describe the solution you'd like**
I would like lynis to detect the package so that it doesn't find an exception.

Thank you so much!
Greetings!

mboelen:
What is the package name of the kernel that you see when running `dpkg -l`?

Then it can be added in include/tests_ports_packages on line 1345
`KERNEL_PKG_NAMES="linux-image-[0-9]|raspberrypi-kernel|pve-kernel-[0-9]"`

Would be great if you can test it, as this is very specific. If it works, a pull request would be very welcome (to give you the credit for reporting and testing).

ferorge:
Hi @mboelen 

The solution you indicate is correct.

Line 1340 of the file /usr/share/lynis/include/tests_ports_packages looked like this:

> KERNEL_PKG_NAMES="linux-image-[0-9]|raspberrypi-kernel|pve-kernel-[0-9]|linux-odroid-5422"

Output from lynis.log:

> 2024-07-23 06:02:15 Performing test ID PKGS-7410 (Count installed kernel packages)
> 2024-07-23 06:02:15 Test: Checking how many kernel packages are installed
> 2024-07-23 06:02:15 Result: found 1 kernel packages on the system, which is fine
> 2024-07-23 06:02:15 ====

The /boot/zImage file is the kernel.
I attach the output of the command:

> dpkg -L linux-odroid-5422
> 
> /.
> /boot
> /boot/exynos5422-odroidhc1.dtb
> /boot/exynos5422-odroidxu3-lite.dtb
[...]

PR #1526, opened 2024-07-23
Fix issues #1496 and #1497

ferorge:
Added inetd package and service name used in ubuntu 24.04. 
Added kernel name for Hardkernel odroid XU4. 

Fernando Orge, ferorge@texto-plano.xyz


mboelen:
Thanks, merged!

Issue #1527, opened 2024-07-24
default_gateway not found


xezpeleta:
**Describe the bug**
Lynis can't detect the default gateway, as Ubuntu 24.04 does not have `netstat` pre-installed. 

**Version**
 - Distribution: Ubuntu 24.04 LTS
 - Lynis version: 3.0.9-1 (installed via apt)

**Expected behavior**
Ideally, default gateway should be retrieved with pre-installed tools (ip command?).
As alternative, lynis debian/ubuntu package should has `netstat` as dependency

**Output**

```
Skipped test NETW-3001 (Find default gateway (route))
```

Thank you for this amazing project!

PR #1528, opened 2024-07-24
Find default_gateway using "ip" instead of "netstat"

xezpeleta:
Hello,

Ubuntu 24.04 LTS does not have netstat installed so Lynis can't detect the default gateway in the system. #1527 

The "ip" command is typically pre-installed on most Linux distributions, so maybe it could be a good alternative.

What do you think?

This patch uses the `ip` command to retrieve the default gateway.

Thanks

### DIFF
--- a/include/tests_networking
+++ b/include/tests_networking
@@ -276,11 +276,11 @@
     # Test        : NETW-3001
     # Description : Find default gateway (route)
     # More info   : BSD: ^default   Linux: 0.0.0.0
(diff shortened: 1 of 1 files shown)


mboelen:
Thanks for the patch (and patience). While the change itself is great, it will break the test completely for systems not having 'ip', like BSD. Can you rewrite it to support both?

mboelen:
Hi @xezpeleta - As there was no response yet, can you have a look at my previous comment?

mboelen:
I'm sorry, but have to close this one without merging. The current pull request can't be merged without modifications.

PR #1536, opened 2024-08-30
HTTP-6660 - Wrap quotes properly for dash and other(?) shells

jpasher-lazor:
The test for TraceEnable in Apache (HTTP-6660) has a comparison that is invalid in Dash (the default for /bin/sh in Debian-based distros). This is due to the way it does the string comparisons to try to compensate for an empty string (are there still shells around that need to do that?).


mboelen:
Thanks, merged!

Issue #1544, opened 2024-09-27
Test KRNL-5622 what about SO without systemd (systemctl) - for example Slackware?


teoberi:
**Describe the bug**
https://github.com/CISOfy/lynis/blob/fb5b808944819a2e0571d3aff3a02ec1c760f0eb/include/tests_kernel#L44

**Version**
 - Distribution: Slackware64-current
 - Lynis version: 3.1.2

**Expected behavior**
If it does not exist, the command should not be run:
`systemctl get-default`

**Output**

> [+] Kernel
> ------------------------------------
> /root/lynis/include/tests_kernel: line 44: get-default: command not found

teoberi:
A possible solution:
`        if [ $([ ! -z ${SYSTEMCTLBINARY} ] && ${SYSTEMCTLBINARY} get-default) ]; then`

teoberi:
Can it be accepted?

mboelen:
Added a slightly different if-statement to allow also for proper logging. See commit https://github.com/CISOfy/lynis/commit/03168113ca7d022785ce5dd6daa4f9af895edd4f

Can you test if it works for you?

teoberi:
With my modification, the runlevel is also displayed in the case without systemd. With your modification, the runlevel display is missing if systemd is missing!
I tested in my version with **-n** instead of **! -z** and it doesn't work!

mboelen:
You are right, your solution might be better suited for a wider range of Linux distributions. Want to create a pull request for it?

teoberi:
I can create a pull request, but I would like to study a little more the two code variants.

teoberi:
After a bit of study, I finally stuck with my solution.
I also studied the documentation and the examples from [here](https://www.geeksforgeeks.org/bash-scripting-how-to-check-if-variable-is-set/), after which I tested the versions with **-n** and **-z** both in Slackware64-current and in Ubuntu 24.04.01.

> Option\Value | non-empty | empty | not declared/not set
> -- | -- | -- | --
> -v | true | true | false
> -z | false | true | true
> -n | true | false | false


The only option that worked for both OSes is **! -z**
I will test everything again tomorrow and I would like to add Slackware to the test case with **who -r** after which I will add a pull request.

PR #1548, opened 2024-10-01
Update tests_kernel -> KRNL-5622

teoberi:
Fix runlevel detection for SO without systemd (eg. Slackware)
Close issue #1544


mboelen:
Merged. The combination within the IF-statement looks slightly complicated, but let's see if things works as intended.

PR #1549, opened 2024-10-01
Update tests_boot_services -> (E)LILO warning

teoberi:
Adding "discontinued" warning for LILO/ELILO.
As of January 2016, LILO is no longer actively developed. ELILO project is orphaned, Debian dropped it in 2014, and RH & SUSE stopped using this tree (and feeding back change) long before that so no longer interested in working on it. Apart from Slackware, which still has packages for LILO and ELILO, I don't know if anyone else uses them, and anyway, Slackware has an active and well-configured package for GRUB.

### DIFF
--- a/include/tests_boot_services
+++ b/include/tests_boot_services
@@ -484,6 +484,7 @@
             else
                 LogText "Result: can not read ${LILOCONFFILE} (no permission)"
             fi
+            ReportWarning "${TEST_NO}" "LILO is discontinued (unmaintained). Recommended alternatives: GRUB."
         else
             LogText "Result: LILO configuration file not found"
         fi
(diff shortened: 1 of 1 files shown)


wileyhy:
From what I understand from Wikipedia, Syslinux is still in use for booting
ISO 9660 disk images.

teoberi:
I have never used Syslinux, but from what I found on the net it seems that it is a bit dead.
https://wiki.syslinux.org/wiki/index.php?title=The_Syslinux_Project
https://git.kernel.org/pub/scm/boot/syslinux/syslinux.git/
https://fedoraproject.org/wiki/Changes/BIOSBootISOWithGrub2

> Modify [lorax-generic-templates](https://github.com/weldr/lorax/tree/master/share/templates.d/99-generic) to use GRUB2 when booting the boot.iso on BIOS systems, instead of syslinux. Upstream syslinux development is dead, and the Fedora maintainer would like to drop the package from the distribution. GRUB2 works as a replacement in most situations and continues to have upstream support.

Issue #1273, opened 2022-02-21
lynis modifies file owner and group


AWHubGit:
I have two different installations with Arch Linux x86_64 5.16.10-arch1-1.
On both is Homebrew for Linux installed with default setup.
[homebrew-on-linux](https://docs.brew.sh/Homebrew-on-Linux#homebrew-on-linux-community)
I tested first Arch Linux with lynis 3.0.7 "sudo lynis audit system".
After that brew didn't work anymore. Fatal error.
I tested second Arch with lynis. Brew also threw "fatal error" after.

Lynis modified/altered owners/groups on dirs and files in path /home/linuxbrew/.linuxbrew.

Why this? Is this wanted?

mboelen:
Lynis is an audit tool that only reviews the system and not applies changes. One possibility is that Brew is making those changes when it is being executed.

To see if this is the case, I suggest the following:
- Restore the permissions
- Run Brew as root (using sudo) and query the installed packages
- See if that changes the permissions as well

AWHubGit:
Brew can't be, shouldn't be run as root:
sudo brew
[sudo] Passwort .....: 
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.

Find it out your self: Install Arch Linux and install Brew for Linux as per defaults. Then install lynis and run it.

On both Arch Linuxes I have with brew for Linux brew didn't run anymore after running lynis with sudo due to changed permissions in path /home/linuxbrew/.linuxbrew.

mboelen:
I have no Arch Linux installed, so can't test that. If you disable the related brew test (most likely that is PKGS-7303) in your custom.prf (skip-test=PKGS-7303), does it then keep working?

mboelen:
Hi @AWHubGit - Did you get the chance to test?

AWHubGit:
I didn't test with skip-test=PKGS-7303 yet. I created an archive with lynis logs inside.
[lynis_logs](https://github.com/AWHubGit/lynis_brew.git)
Rapidly parsing the logs seen that lynis seems to harden:
2022-02-26 11:53:47 Hardening strength: System has been hardened, but could use additional hardening
Does this change owners/permissions?
SELinux is a hardened fs but not officially supported by Arch Linux.

xnoguer:
@mboelen @AWHubGit I couldn't replicate the behaviour described using an archlinux:latest Docker container.
[...]

PR #1564, opened 2024-10-21
Run brew commands as regular user when Lynis is invoked as root

### LINKED ISSUE #382: Lynis needs to drop privilleges to support homebrew on MacOSX
Given I run the version of today + mac ports patch:
When scanning I need to login as root.
Then I get the following error in my report:
> 
> [+] Ports and packages
> ------------------------------------
>   - Searching package managers
>     - Searching brew                                          [ FOUND ]
>     - Querying brew for installed packages
> Error: Running Homebrew as root is extremely dangerous and no longer supported.
> As Homebrew does not drop privileges on installation you would be giving all
> build scripts full access to your system.
>   - Checking ports for updates                                [ FOUND ]

Logfile /var/log/lynis.log doesn't give any more details.
Expected behavior: the program should drop privileges.

Neved4:
Runs Homebrew commands as original user when Lynis is invoked as root

Fixes #382, related #1273.

Instead of running Homebrew as the root user, this change runs it as a normal user, which is the intended use case for Homebrew. After the change, the command is run a non-root user context, which seems a way to effectively drop privileges.

> [!NOTE]  
> The function could have a different or more descriptive name, refactored to be more compact, etc. See:
> ```sh
> RunBrewAsUser() {
>     { "$(id -u)" && sudo -u "$SUDO_USER" brew "$@"; } || brew "$@"
> }
> ```

***

## Security

Thanks to `brew`'s safeguards this might not be exploitable, but including the following information for reference.

MITRE [CWE-250] is related to this change:
| CWE ID  | CWE Title                                            | Description                                                                                                      |
|---------|-----------------------------------------------------|------------------------------------------------------------------------------------------------------------------|
| [CWE-250] | Execution with Unnecessary Privileges                | The software runs with more privileges than necessary, which can lead to security vulnerabilities.               |

This change follows [REF-76] by dropping the privileges as soon as possible:

> _Run your code using the lowest privileges that are required to accomplish the necessary tasks._
> _Wrap and centralize this functionality if possible, and isolate the privileged code as much as possible from other code._
> _Raise privileges as late as possible, and drop them as soon as possible to avoid [CWE-271]._  
>
> Source: [_CWE-250: Execution with Unnecessary Privileges_][CWE-250].

[CWE-250]: https://cwe.mitre.org/data/definitions/250.html
[CWE-271]: https://cwe.mitre.org/data/definitions/271.html
[REF-76]: https://web.archive.org/web/20211209014121/https://www.cisa.gov/uscert/bsi/articles/knowledge/principles/least-privilege


wileyhy:
So this function:

RunBrewAsUser() {
    { "$(id -u)" && sudo -u "$SUDO_USER" brew "$@"; } || brew "$@"
}

...begins by expanding the command substitution, `"$(id -u)"`. In Bash 5.2
in Posix mode, this comsub outputs '10151' to Stdout. On a Linux machine it
might print something in the low 1000's range. Any code before '&&' is read
by the shell as one entire command, so, since this particular command line
contains just one string, the shell will most likely search PATH for a
command named 10151 (unless there is an alias or function by that name).
This is probably not what was intended.

In Linux, system commands can begin with digits, but neither 10151 nor 1000
nor 0 are common command names, so a bad actor could place a malicious
script or program in PATH which the kernel would then attempt to execute.

If lynis had been executed with root permissions, and if such a malicious
script or program as this existed, then, my understanding at this time is
that the kernel would be instructed by the user to execute such a program
with root privileges. :-\
Wiley

wileyhy:
Have I gotten anything wrong in this reply? It's been kind of a weird day.

Neved4:
Hey @wileyhy thx for looking into this! You're 100% right for noting it.

Additionally:
1. The function above is not the one that is in the commit.
2. You're right `$(id -u)` evaluates as-is instead of being compared to anything there.
3. This is a typo. It should have had the `[ "$(id -u)" -eq 0 ]` comparison.

We could edit my post to update this suggestion, or remove it since it doesn't reflect the current change.

Cheers!

wileyhy:
Hi @neveda,

  If Homebrew is installed on a system, then perhaps util-linux could also
be installed from \brew\?  That might allow for dropping privileges by use
of \runuser\, \setpriv\ or \su\, assuming Homebrew's version has any of
those binaries.

  I don't know for certain at the moment, but dropping privileges in a
strict POSIX _shell_ context might not be possible. C17 is specified by
POSIX, but how else could it be done?

Wiley

PR #1587, opened 2024-12-17
merge upstream

gcsgithub:


mboelen:
I see that you opened and closed this. No action required from the Lynis project, right?

PR #1589, opened 2024-12-18
Fix #1562

### LINKED ISSUE #1562: kadmin.local: unable to get default realm
**Describe the bug**
On Fedora 40 the new (312) version of lynis outputs the message,

        kadmin.local: unable to get default realm

This occurs during the Kerberos section but is not affected by "2>&1" redirection,
so goes to the "terminal" or preceding/calling process output (eg. in my case cron).
I assume it's coming directly from the kadmin.local binary so there probably needs
to be a test of krb config before this is called & not called at all under conditions
where krb is not used.
However I'm sure there could be circumstances where a bad actor might hide a krb
config & potentially use it for secure channels/auth.

**Version**
 - Distribution Fedora 40 (fc40)
 - Lynis version 3.1.2-1.fc40

**Expected behavior**
The (error?) message should be appropriately directed, & able to be redirected, when
kadmin.local is being called from within lynis.

In my case where there is no krb config it should probably not occur.
ie. an attempt to get the "default realm" will fail & so probably should not even be
attempted.

**Output**
 kadmin.local: unable to get default realm

**Additional context**
Followed the 312 version install on FC40 after the "grep usage" messages were fixed.

pyllyukko:
kadmin.local binary might exist, even though Kerberos is not configured and /etc/krb5.conf does not exist.

### DIFF
--- a/include/tests_kerberos
+++ b/include/tests_kerberos
@@ -14,7 +14,7 @@ InsertSection "${SECTION_KERBEROS}"
         PREQS_MET="YES"
         # Make sure krb5 debugging doesn't mess up the output
         unset KRB5_TRACE
-        PRINCS="$(${KADMINLOCALBINARY} listprincs | ${TRBINARY:-tr} '\n' ' ')"
+        PRINCS="$(${KADMINLOCALBINARY} listprincs 2>/dev/null | ${TRBINARY:-tr} '\n' ' ')"
         if [ -z "${PRINCS}" ]
         then
             PREQS_MET="NO"


jwadodson:
OK, so what are the conditions where that might occur?
Would not the best option to test it's existence before throwing such an error?
& then trying to be more specific about why  the kadmin.local binary might exist
when other conditions for it's safe & usual operation do not exist?

Is the binary ever installed "without control"?
Under what special circumstances?
What are the dangers?
etc.?

Cheers

John

jwadodson:
I have now been running lynis.noarch 3.1.3-1.fc41 for a few days.

It still gives this output (one line)

        kadmin.local: unable to get default realm

if I run lynis using this command line...

       lynis audit system 2>&1 > config/chklynis.$DATE

If that error was able to be redirected to my "log" file, I'd be satisfied, if not happy
But an error that is unable to be redirected for later analysis/processing is a little annoying.

As I said previously I think I have no trace of kerberos on this system.

pyllyukko:
> OK, so what are the conditions where that might occur? Would not the best option to test it's existence before throwing such an error? & then trying to be more specific about why the kadmin.local binary might exist when other conditions for it's safe & usual operation do not exist?
> 
> Is the binary ever installed "without control"? Under what special circumstances? What are the dangers? etc.?

Clearly you have the `kadmin.local` binary as it throws an error. You can also see it in Lynis' logs (`lynis.log`), e.g.:

> Found known binary: kadmin.local (krb5) - /usr/sbin/kadmin.local

You can see the actual code that discovers the binary [here](https://github.com/CISOfy/lynis/blob/f2ac59a1303f397e263404d10f1315977372b58e/include/binaries#L199). You probably have some Kerberos package installed in your system, but this is really something you need to look into on your own.

The fact that you get an error from `kadmin.local` was an oversight from me when I initially implemented this functionality (see #1456). And this fix would just discard the error in the prerequisites test.

> I have now been running lynis.noarch 3.1.3-1.fc41 for a few days.

It is not fixed yet in that version, as the potential fix resides in this particular pull request (which has not been merged yet) and I would ask you to test this fix and see if it resolves your issue?

jwadodson:
OK thanks.
/usr/sbin/kadmin.local was installed as part of krb5-server which was called for by samba-dc
which is not being used.
Hopefully that will fix it for me.
I'll leave it till the lynis run tonight to check...

pyllyukko:
> Hopefully that will fix it for me.
> I'll leave it till the lynis run tonight to check...

No. The problem will not fix itself. You will need to manually try out the [proposed fix](https://github.com/CISOfy/lynis/pull/1589/files) in this PR and report back whether it worked.

jwadodson:
I have removed  krb5-server & samba-dc so I should not now see that error, as you thought it was coming
from the /usr/sbin/kadmin.local binary? With /usr/sbin/kadmin.local  gone is that not the case?

pyllyukko:
> I have removed krb5-server & samba-dc so I should not now see that error, as you thought it was coming
> from the /usr/sbin/kadmin.local binary? With /usr/sbin/kadmin.local gone is that not the case?

Well it mitigates the problem for you when the `kadmin.local` binary is gone, but it will happen to someone else later on. So in order to fix it in Lynis, I again ask you to try this fix in such a way, that the `kadmin.local` binary is still there.

jwadodson:
It looks like it will work (stderr > /dev/null) & having reinstalled krb5-server on it's own (which supplies
the /usr/sbin/kadmin.local binary) it works as expected (ie. no error from the  /usr/sbin/kadmin.local binary)

pyllyukko:
E.g.:

1. Git clone Lynis with `git clone https://github.com/CISOfy/lynis.git`
2. Fetch this PR with `git fetch origin pull/1589/head:fix-1562`
3. Switch to the proper branch `git switch fix-1562`
4. Run Lynis from there (and not the one installed on your OS) `./lynis audit system --profile developer.prf --tests-from-group kerberos --no-plugins --usecwd`

pyllyukko:
> It looks like it will work (stderr > /dev/null) & having reinstalled krb5-server on it's own (which supplies
> the /usr/sbin/kadmin.local binary) it works as expected (ie. no error from the /usr/sbin/kadmin.local binary)

Looks like, or did you test it? Can you please try with the instructions provided in my [previous comment](https://github.com/CISOfy/lynis/pull/1589#issuecomment-2565155128) and report back?

jwadodson:
I did it by adding that redirect to that line (17) of /usr/share/lynis/include/tests_kerberos

Then running lynis.
It worked as expected & I did not get that error.

Sorry but I really can't get involved in playing "git". Having got a 2nd new hip almost 2 weeks ago, I just
can't sit at a computer for long enough at the moment.

mboelen:
Thanks, merged!

PR #1592, opened 2025-01-14
Report a warning if no logrotate configuration is found

pSub:
Previously the text of the report would suggest a warning but report it as a suggestions. This was an inconsistency in the reporting as a warning was displayed in the report, but the summary showed

    Great, no warnings

in case no other warning was reported.


pSub:
Is there anything preventing this from being merged? If so, I am more than happy to resolve it.

pSub:
Understood. Thanks. I updated the commit accordingly.

mboelen:
Merged, thanks!

Issue #1600, opened 2025-02-08
Basic support for OpenWrt


macie:
This issue collects items needed for basic support of OpenWrt. The goal is to run `lynis` on `OpenWrt` without errors.

- [ ] OS detection + EOL check
- [ ] hostname detection
- [ ] `HOSTID` creation (fix `GetHostID` exceptions)
- [ ] fix `Unknown service manager` exception
- [ ] support BusyBox `ps` (fix `/bin/ps: unrecognized option: x` and `/bin/ps: unrecognized option: a` errors)

(the list may grow with time)

graysky2:
@macie - can you update with https://github.com/CISOfy/lynis/pull/1748

PR #1604, opened 2025-02-09
[OpenWrt] Fix HostID and HostID2 creation exceptions

macie:
This is a part of enabling support for OpenWrt, see: https://github.com/CISOfy/lynis/issues/1600.

(more details in commit messages)


mboelen:
Looks good to me. My compliments for how you extended the code with careful attention on the existing parts.

Issue #1597, opened 2025-01-22
Unknown OS found in /etc/os-release: OpenMandriva


bcoles:
```
  Exception found!

  Function/test:  [OS Detection]
  Message:        Unknown OS found in /etc/os-release - Please create an issue on GitHub and share the contents (cat /etc/os-release): https://github.com/CISOfy/lynis
```

```
NAME="OpenMandriva Lx"
VERSION="5.0 (Iodine) Rock"
ID="openmandriva"
VERSION_ID="5.0"
PRETTY_NAME="OpenMandriva Lx 5.0 (Iodine) Rock"
BUILD_ID="20231104.16"
VERSION_CODENAME="iodine"
ANSI_COLOR="1;43"
LOGO="openmandriva"
CPE_NAME="cpe:/o:openmandriva:openmandriva_lx:5.0"
HOME_URL="http://openmandriva.org/"
BUG_REPORT_URL="https://github.com/OpenMandrivaAssociation/distribution/issues/"
SUPPORT_URL="https://forum.openmandriva.org"
PRIVACY_POLICY_URL="https://www.openmandriva.org/tos"
```

mboelen:
Thanks for reporting. Should be resolved in next release with the pull request of @jpasher-lazor

PR #1608, opened 2025-02-25
Add detection for OpenMandriva Lx

jpasher-lazor:
This adds detection for OpenMandriva Lx as reported in #1597


mboelen:
Thanks, merged!

PR #1648, opened 2025-07-22
Pull password length from pam_passwordqc

puck:
Also return passphrase settings if configured. People may want to enforce these settings differently from password settings.


puck:
I happy to provide a sample PAM config file for pam_passwordqc if that'd be useful.

PR #1656, opened 2025-08-06
KRNL-5820: Allow whitespaces immediately before or after the "="

pyllyukko:
INI files can have whitespaces around the "=" and Lynis should be able to parse those too.

https://www.freedesktop.org/software/systemd/man/latest/systemd.syntax.html: "Whitespace immediately before or after the "=" is ignored."


samanthagrayxoxo-afk:
What’s though

PR #1679, opened 2025-11-07
Add Fedora 43 EOL date

pointydev:
Fedora 43 recently hit public release last month, and as such I've added it's EOL date to the database. I've also added a link to the Fedora project schedule information site, which lists EOLs for upcoming and currently supported releases (although these are subject to change until the N+2 release is completed).

Thanks,
Elliott


mboelen:
Thanks, merged!

PR #1681, opened 2025-11-14
Add Linux Mint 22.2 EOL

Skorium:
Hi,

Another End Of Life date PR for another Linux distro I'm using. As per Linux Mint page: https://www.linuxmint.com/rel_zara.php

Skorium


mboelen:
Thanks for the pull request. Why did you close/remove your patch? There is no patch available now, so I can't merge it ;-)

PR #1682, opened 2025-11-16
Add openSUSE Leap to EOL DB

mnhauke:
This PR adds openSUSE Leap version 15 and 16 to the EOL DB.

Information gathered from https://en.opensuse.org/Lifetime


mboelen:
Thanks, merged!

PR #1683, opened 2025-11-16
Fix OS detection for CachyOS (ID=cachyos)

cycloarcane:
Lynis currently throws "Unknown OS found in /etc/os-release" on CachyOS.

Cause:
`/etc/os-release` defines `ID=cachyos` (all lowercase), but the osdetection
case block uses `"CachyOS")`, which does not match due to case sensitivity.

Fix:
Adjusted the case label to `"cachyos")` to correctly match the ID value.
Verified locally with:

    sudo ./lynis audit system --usecwd

Resulting output includes:
    Operating system name:     CachyOS
    Operating system version:  Rolling release

This resolves the OS detection exception.
Screenshot_20251116_193048


khodekia:
looks like I was late  to the party.

mboelen:
No worries @khodekia, most importantly is that it works :-)

PR #1688, opened 2025-11-21
Update RHEL 8 EOL and add RHEL 9 and 10 dates to software-eol.db

avegarten:
While running lynis on my RHEL 10.1 laptop recently and noticed the following message: 

"Notice: No OS entry was found in the end-of-life database"

After inspecting software-eol.db, I found that RHEL 9 and 10 entries were missing entirely, so this PR adds their corresponding EOL dates. While adding the entries for RHEL 9 and 10, I also noticed that the entry for RHEL 8 did not match the Red Hat offical policy already listed for AlmaLinux/Rocky Linux 8. So, I have changed it to match.


mboelen:
Thanks, merged!

PR #1702, opened 2025-12-31
Add Alpine 3.23 EOL date

0xiod:
image
I have noticed that there is no EOL date for new Alpine Linux release.


mboelen:
Thanks, merged!

PR #1714, opened 2026-01-21
Add better NixOS detection

Pandapip1:
Because it's funny, I have my /etc/os-release customized to use a `NAME` that's not `NixOS`. This is something that's easy to do by simply setting the options that control the stuff in os-release:

https://github.com/NixOS/nixpkgs/blob/527be1432109cfaf305987865bc17cc908e8e77a/nixos/modules/misc/version.nix#L121-L161

In fact, there's an entire library dedicated to making silly customizations like this: https://github.com/yunfachi/denix

This changes the NixOS detection code to look for /run/current-system, which is present on all NixOS systems regardless of configuration, and is importantly _not_ present when using Nix on non-NixOS.

### DIFF
--- a/include/osdetection
+++ b/include/osdetection
@@ -382,12 +382,6 @@
                             OS_REDHAT_OR_CLONE=1
                             OS_VERSION=$(grep "^VERSION_ID=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
                         ;;
-                        "nixos")
-                            LINUX_VERSION="NixOS"
-                            OS_NAME="NixOS"
-                            OS_VERSION=$(grep "^VERSION_ID=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
-                            OS_VERSION_FULL=$(grep "^VERSION=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
-                        ;;
                         "ol")
                             LINUX_VERSION="Oracle Linux"
                             OS_NAME="Oracle Linux"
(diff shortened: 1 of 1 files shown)


Pandapip1:
Bump!

mboelen:
Thanks for the pull request.

This change introduces a new variable (ETC_RELEASE_DETECT_FAILED), but does not add any functionality. Also there is a risk that the path /run/current-system may be used by other distributions in the future, so how do we then know it's NixOS or something else?

Pandapip1:
> This change introduces a new variable (ETC_RELEASE_DETECT_FAILED), but does not add any functionality. 

The existing code assumes that if `/etc/os-release` exists, then if it's not in the giant switch statement, that it's an unknown distro. Therefore, that code path needs to be bypassed.

> Also there is a risk that the path /run/current-system may be used by other distributions in the future, so how do we then know it's NixOS or something else?

For the time being, NixOS is the only one that uses this path. Also, it's at the very end; if the distro has already been detected it won't override it. If this causes a false positive for some distro that isn't even recognized by lynis in the first place at some point in the future, that can be dealt with; a theoretical false positive edge case is probably better than current NixOS systems being not detected as NixOS.

mboelen:
I'm worrying about the long-term complexity that this edge-case introduces. After all, if you decide to make your OS show up as a non-default one, then you risk of it not being detected properly by tools.

I rather see that we then use the ID_LIKE as a backup.

Pandapip1:
Didn't notice the `ID_LIKE`; that definitely seems like the best solution. I'll add this to my backlog.

PR #1740, opened 2026-05-07
Fix NETW-2706: Update DNSSEC detection logic for modern resolvectl

miikun95:
<section>
  <h2>summary</h2>
  <p>this pr improves the dnssec detection logic in the `netw-2706` test to support newer versions of `systemd-resolved` (systemd 250 and later).</p>
</section>
<section>
  <h2>problem</h2>
  <p>the current implementation relies on `resolvectl --statistics`, which only provides statistical counters and does not reliably indicate whether dnssec is active for the current network interfaces. furthermore, the previous `awk` parsing logic fails on newer output formats like `dnssec=yes/supported`.</p>
</section>
<section>
  <h2>changes</h2>
  <ul>
    <li>changed `resolve_cmd_param` from `--statistics` to `status` to get the actual operational status.</li>
    <li>updated the parsing logic using `grep` and `cut` to correctly extract the status even when the output includes additional info (e.g., `yes/supported`).</li>
    <li>added `head -n 1` to ensure a consistent single-line result when multiple interfaces are active.</li>
  </ul>
</section>
<section>
  <h2>note for maintainers</h2>
  <p>i am a japanese user and am using a machine translator to write this. i apologize for any unnatural phrasing. i have verified this fix on my local environment, and it now correctly returns `yes`.</p>
</section>
<section>
  <h2>verified result</h2>
  <dl>
    <dt><b>command:</b></dt>&nbsp;<dd><pre><code>resolvectl status | grep -o 'DNSSEC=[^ ]*' | cut -d'=' -f2 | cut -d'/' -f1 | head -n 1</code></pre></dd>
    <dt><b>Output:</b></dt>&nbsp;<dd>yes</dd>
  </dl>
</section>
<aside>
  <p>miikun95, miikun95@miikun95.net</p>
</aside>

### DIFF
--- a/include/tests_networking
+++ b/include/tests_networking
@@ -244,7 +244,7 @@
     if [ -n "${RESOLVECTLBINARY}" ]; then
         PREQS_MET="YES"
         RESOLVE_CMD="${RESOLVECTLBINARY}"
-        RESOLVE_CMD_PARAM="statistics"
+        RESOLVE_CMD_PARAM="status"
     elif [ -n "$(command -v systemd-resolve 2> /dev/null)" ]; then
         PREQS_MET="YES"
         RESOLVE_CMD="$(command -v systemd-resolve 2> /dev/null)"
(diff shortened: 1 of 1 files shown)


q1sh101:
I tested the `statistics` to `status` change on Lynis 3.1.6, Ubuntu 26.04,
and systemd 259. On this system it also fixes a second problem in NETW-2706.

As an unprivileged user, `resolvectl statistics` requests the polkit action
`org.freedesktop.resolve1.dump-statistics`. The local policy sets
`allow_active` to `auth_admin_keep`, so the command opens an authorization
dialog and waits. Closing the dialog leaves the test UNKNOWN.

Measured locally:

```console
$ resolvectl statistics >/dev/null
# opens a polkit dialog and waits

$ resolvectl status >/dev/null
# exits 0 immediately, no polkit authorization event
```

`status` still carries the value this PR parses:

```text
Protocols: -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
```

So on this configuration the change fixes both the DNSSEC parsing and an
unexpected interaction. It may be worth mentioning the second effect in the
commit message.

PR #1744, opened 2026-05-19
Update software-eol.db for OpenBSD expiry dates.

icepic:
OpenBSD supports two releases, so dates for 7.8 and 7.9 are set to 6 or 12 months in the future respectively, a release schedule that has held for some 50+ releases in a row.


mboelen:
Great, merged!

Thanks for your contribution, it will be part of the next release.

PR #1755, opened 2026-06-16
Feat improve busybox compatibility

bravl:
This PR improves the overall compatibility with Busybox based systems.
- Align parameters of "stat" to use "-c" instead of "--format". "--format" is not supported by busybox "stat"
- Replace `-printf` with a grep based solution to which works on both busybox and none-busybox systems
- Improve busybox os detection, solving an edge case if "/bin/" is a symlink to "/usr/bin".

Tested on an upstream buildroot Qemu build.


wileyhy:
Hi,

- the parameter expansion "${sFILE%/*}", which is Posix-2008 compliant,
could replace `$(dirname ${sFILE}`, so a fork()/exec() could be eliminated.

- `which busybox` is executed three times when it's only necessary to
execute it once. Duplicate commands performing identical fork()/exec()
procedures slow the program somewhat. I'd start with BUSYBOX_PATH and use
the output of that for the other two.

Wiley

Issue #1352, opened 2022-12-12
Changes needed in AUTH-9230


borislavba:
SHA_CRYPT_{MIN,MAX}_ROUNDS are checked when ENCRYPT_METHOD is not SHA256/512 (e.g. Fedora 37's default is Yescrypt). According to the comments in /etc/login.defs they are used only when ENCRYPT_METHOD is SHA256/512.
Maybe the checks should be rewritten according to the ENCRYPT_METHOD selected.

Also note this:
In Fedora's login.defs there's this: "Currently SHA_CRYPT_MIN_ROUNDS is not supported"
In ArchLinux there is no such comment...
So I'm a bit confused if the SHA_CRYPT_MIN_ROUNDS should be checked or not.

 - Distribution Fedora 37, ArchLinux (shadow 4.12.3 in both)
 - Lynis version 3.0.8

vk6xebec:
Make sure the MIN and MAX ROUNDS are uncommented. Lynis will not throw an error. The system will ignore those fields.

MFTabriz:
> Make sure the MIN and MAX ROUNDS are uncommented. Lynis will not throw an error. The system will ignore those fields.

I thought it goes without saying: Lynis should not propose changes which will be ignored by system.

beatquantum:
Debian 13 Trixie is no longer using SHA512 rounds and the default is yescrypt.
In such cases, this check may not be essential.

PR #1760, opened 2026-07-01
AUTH-9230: skip SHA_CRYPT rounds check when ENCRYPT_METHOD is YESCRYPT

adjenk:
Related to #1352

## Problem
AUTH-9230 checks SHA_CRYPT_MIN_ROUNDS/SHA_CRYPT_MAX_ROUNDS in
/etc/login.defs, but these settings only apply when ENCRYPT_METHOD
is SHA256 or SHA512. On systems where ENCRYPT_METHOD is set to
YESCRYPT (the default on Fedora 35+, Debian 11+, Arch Linux, and
Ubuntu 22.04+), these variables are correctly absent, but the test
still reports password hashing rounds as unconfigured and issues a
suggestion; a false positive.

## Fix
Adds a check on ENCRYPT_METHOD in /etc/login.defs. When it's set to
YESCRYPT, the SHA_CRYPT rounds check is skipped entirely rather than
reporting a false "disabled" result. 

## Scope
This addresses the yescrypt false-positive described in #1352, but
not the separate question raised in that issue about whether
SHA_CRYPT_MIN_ROUNDS is actually honoured on some distros.
This PR does not check YESCRYPT_COST_FACTOR (yescrypt's tuning
parameter). Whether that needs its own check is a separate question.

## Testing
Verified locally with `lynis audit system --tests AUTH-9230 --debug`
against a login.defs with ENCRYPT_METHOD YESCRYPT (test now skips
correctly).


mboelen:
I see you closed deleted your branch, so I can't merge your pull request :cry: 

Can you recreate it @adjenk ?

adjenk:
Recreated as #1772

PR #1764, opened 2026-07-16
Update software-eol.db with FreeBSD releases

kravietz:
Based on https://www.freebsd.org/security/#sup


mboelen:
Thanks! This will be available in next release.

PR #1772, opened 2026-08-09
AUTH-9230: skip SHA_CRYPT rounds check when ENCRYPT_METHOD is YESCRYPT

adjenk:
## Problem
AUTH-9230 checks SHA_CRYPT_MIN_ROUNDS/SHA_CRYPT_MAX_ROUNDS in
/etc/login.defs, but these settings only apply when ENCRYPT_METHOD
is SHA256 or SHA512. On systems where ENCRYPT_METHOD is set to
YESCRYPT (the default on Fedora 35+, Debian 11+, Arch Linux, and
Ubuntu 22.04+), these variables are correctly absent, but the test
still reports password hashing rounds as unconfigured and issues a
suggestion; a false positive.

## Fix
Adds a check on ENCRYPT_METHOD in /etc/login.defs. When it's set to
YESCRYPT, the SHA_CRYPT rounds check is skipped entirely rather than
reporting a false "disabled" result.

## Scope
This addresses the yescrypt false-positive described in https://github.com/CISOfy/lynis/issues/1352, but
not the separate question raised in that issue about whether
SHA_CRYPT_MIN_ROUNDS is actually honoured on some distros.
This PR does not check YESCRYPT_COST_FACTOR (yescrypt's tuning
parameter). Whether that needs its own check is a separate question.

## Testing
Verified locally with lynis audit system --tests AUTH-9230 --debug
against a login.defs with ENCRYPT_METHOD YESCRYPT (test now skips
correctly).

Recreates #1760, which was closed when the fork was deleted.


mboelen:
Bummer, I missed it when you recreated it, so can't merge it. Do you have automatic deletion enabled?

adjenk:
> Bummer, I missed it when you recreated it, so can't merge it. Do you have automatic deletion enabled?

Sorry! No, I just keep trying to clean up repos, perhaps a bit too often sometimes.

Recreated as #1777

PR #1784, opened 2026-09-14
Add support for 'poky-altcfg' in OS detection

Noightmore:
There is a variation of the Poky distro -- poky-altcfg which upon detection is marked as unknown distro. I added detection approach that is the same to the way "manjaro" and "manjaro-arm" are handled.

Output of cat /etc/os-release:
ID=poky-altcfg
NAME="Poky (Yocto Project Reference Distro)"
VERSION="6.0.3 (wrynose)"
VERSION_ID=6.0.3
VERSION_CODENAME="wrynose"
PRETTY_NAME="Poky (Yocto Project Reference Distro) 6.0.3 (wrynose)"
CPE_NAME="cpe:/o:openembedded:poky-altcfg:6.0.3"


mboelen:
Looks good, thanks!

